PR #13087, opened 2017-02-24
Adding stan's tickets from TG

cmungall:
This was the last set before we temporarily switch off TG


ukemi (approved):
This looks good. Let's merge.

PR #13110, opened 2017-03-01
Glut neuron diff 13107

krchristie:


ukemi (commented):
Hi Karen,

If you delete the asserted is_a parents in your stanzas and run the reasoner, you will see that these terms will have inferred is_a parents to the ones you assert. There is no longer any need to assert them if they are inferred. They will be added appropriately, but it is always a good idea to see what is inferred and that it makes sense. So if that all works, you can go in and delete your asserted is_a parents. If this doesn't make sense, we can talk about it.

krchristie:
Maybe we should talk because I have some questions.

Do you have time now?

-K

ukemi:
Yup.

Issue #10873, opened 2014-03-03
Mass addition of Metacyc reactions to MF branch


gocentral:
This request came in to myself (Tanya) and Donghui. I'm opening a SF ticket to track progress.

There are about 1700 metacyc reactions that are not on the metacyc2go mapping file. We would like to have the equivalent GO terms created for them so they can be included in the mapping file. In the end we would like to submit GO annotations for 12 plant genomes using the updated mapping file.

For each of the 1700 metacyc reactions, there is a 4-part or 3-part EC which can be used for assigning to a proper parent term. We could also provide reaction equation, and suggested term name. In the past, many yeas ago, Amelia was my contact person for updating the mapping file. Who should I contact now? Please advise.

Thank you,
Peifen

Peifen = Peifen Zhang @ PlantCyc

Reported by: tberardini

Original Ticket: [geneontology/ontology-requests/10687](https://sourceforge.net/p/geneontology/ontology-requests/10687)

ukemi:
The plan forward is to have Sue's group (Peifen) create obo stanzas for the new molecular function terms that will be sent for review by ontology curators (David and Harold). We will set aside some time each week to review these similar to the way we did the chebi project, and as we approve them we will cut and paste the stanzas in the ontology file. I have reserved the following IDs for this project:

+GO:0102001	GO:0104000	GO	Molecular Functions for Sue Rhee

ukemi:
Sent Peifen a sample stanza to populate. She will send back a few for Harold and I to look at.

hdrabkin:
Yes, these are being added bit by bit; I am focusing on the low hanging fruit.

hdrabkin:
I am adding 10 terms through git today

hdrabkin:
As of today, after Peifen reviews a set I sent, there are 42 terms left.

hdrabkin:
This is essentially finished; any further tweaking can be done in new tickets.

PR #13116, opened 2017-03-02
NO SEMANTIC DIFFERENCE: this is a troundtrip commit to restore canoni…

cmungall:
…cal ordering


ukemi (approved):
Hmm. These are all enzymes going in as part of #10873. They are added to the end of the file and then run through a load and save cycle. Do they not get re-ordered in the round trip?

hdrabkin:
So these look like they are from what I added yesterday but I saved out of Protege; shouldn't that have re-ordered them.
My workflow: add several stanzas at end of file ahead of first [Typedef] stanza after last real go term, using vi.  Save, then load into protege. Do a little checking that things are ok, then save file to my branch, then commit... then push;

hdrabkin:
So I just did a load and save as from my copy from yesterday and it resorted; I could have sworn I did a save from protege

PR #13123, opened 2017-03-03
three more enzymes

hdrabkin:
Getting a workflow down for adding terms in bulk, as per ticket 10873 . 
So yesterday, Chris & I figured that Protege was not actually saving the file edited by vi upon round-trip attempt (no popup save progress screen, AND the time stamp of the file did not change). So to force I save, I routinely add a space in one of the defs; save; then remove it and save again.  The time stamp changes, and a git diff shows the terms in their correctly sorted place, and not at the end of the file.


ukemi:
Interesting I am adding the documentation label to this so that we can make a note of it if we have to hand-edit the file. So as we saw yesterday, at least one edit is needed in Protege in order for it to actually save a new version of the file. @vanaukenk

hdrabkin:
Yes, and I think it's actually a Protege bug. In any other program that I use, if I do a file save, whether or not I have changed the file, the program goes and saves it (can tell by the time stamp changing.

dosumis:
Given the Protege bug, it would be much more efficient to use robot to reserialize.  Maybe add this to doc?

hdrabkin:
Not familiar with "robot'

dosumis:
> Not familiar with "robot'

Command line tool.  Replacement for owltools (which you could also use).  This would probably work 

owltools go-edit.obo -o -f obo go-edit.obo  

or perhaps safer:

owltools go-edit.obo -o -f obo tmp.obo | mv tmp.obo go-edit.obo

cmungall:
> Yes, and I think it's actually a Protege bug. In any other program that I use, if I do a file save, whether or not I have changed the file, the program goes and saves it (can tell by the time stamp changing.

This isn't a bug. Protege is behaving like every single other Desktop application I've ever used (with the exception of maybe OE). Standard behavior is not to alter file on Save if contents unaltered. You can use "save as" to force a save, but I would avoid this and just do the roundtrip on the command line as David suggests.

hdrabkin:
Maybe it is OE I'm thinking of;

PR #13124, opened 2017-03-03
response to other organism, changing is-a multiorg process to part-of

### LINKED ISSUE #12975: response to other organism
Alice (@zebrafishembryo ) notes that 'response to other organism' (GO:0051707) is [no longer a subclass of multi-organism process](http://www.ebi.ac.uk/QuickGO/GTerm?id=GO:0051707#term=history).  This is causing problems for annotators as the need to be able to record the taxon of the 'other' organism in these cases, but this is only allowed for multi-organism processes.

The change may have come from me (although I haven't manage to track it down).  

The might potentially be justified by saying that it covers the process in the responding organism only, but I wouldn't want to be dogmatic about this. 

So, I propose to change back - but this leaves at least one class as a child of both single an multi-organism process: viral induced premature senescence (GO:0090401)

There is (currently) no axiom declaring disjointness between single and multi-organism processes, so this will not cause the build to fail.  But the disjointness probably should be declared.

After discussion with Alice (@zebrafishembryo) I came up with a possible fix:

Rename -> viral-induction of premature senescence
Add logical def: 'modulation by virus of host process' and 'positively regulates' some 'cellular senescence'

But on reflection, I'm not sure this works as it may be an endogenous response to viral infection, rather than something the virus has evolved to induce:

definition "A cellular senescence process associated with the dismantling of a cell as a response to viral infection."^^xsd:string

Not sure of solution.

cmungall:
Fixes #12975


pgaudet:
already done.

PR #13134, opened 2017-03-07
added cell cycle terms to do not annotate subsets. Fixes #13313

ukemi:


pgaudet (changes requested):
Hi David, 

1. I thought we would add 'DO NOT ANNOTATE' (not 'do not manually annotat') to all (InterPro has even fixed their annotations, see https://github.com/geneontology/go-annotation/issues/1538). 

2. I am not sure this is a big deal but the issue # should have been 13113 (not 13313), so if you used the 'Fixes # to close the issue it will not work. 

Thanks, 
Pascale

PR #13149, opened 2017-03-10
added KICSTOR complex Fixes #13129

### LINKED ISSUE #13129: NTR : KICSTOR complex
Dear GO editors,

Wolfson et al., (PubMed=28199306) describe a new protein complex the KICSTOR complex composed of KPTN, ITFG2, C12orf66 and SZT2 that regulates the TORC1 signaling pathway in response to nutrients.
Could you consider creating a term for this complex.
Thanks in advance,
Lionel

pgaudet:


ukemi (changes requested):
The logical definition is not necessary and sufficient. There are lots of macromolecular complexes that are capable of intracellular signal transduction. So you need to remove the equivalence axioms and make those asserted relationships. I don't even think making this a macromolecular complex that regulates the TORC1 signaling pathway will work in this case, because I assume there are a lot of those as well.

ukemi (approved):
Looks good now.

ukemi:
Arhgh. It should be capable of part of, not just capable of. Sorry.

PR #13152, opened 2017-03-10
added never_in_taxon constraints to term cilium movement

krchristie:


ukemi (approved):
These all look ok to me.

Issue #13137, opened 2017-03-08
obsoletion: GO:1900401 regulation of meiosis by regulation of transcription from RNA polymerase II promoter


ValWood:
obsoletion: GO:1900401 regulation of meiosis by regulation of transcription from RNA polymerase II promoter
and descendants

During previous changes this was re-defined as regulation of "meiotic cell division",
but the paper  which is used to define the term is not for the "regulation of meiotic division", it is 
"regulation of meiotic cell cycle"

I'm not sure that meiotic division is regulated transcriptionally...(it's causally upstream).

experimental annotations 
PomBase 5
SGD 2

these annotations should be rehoused under:
GO:0010672 regulation of transcription from RNA polymerase II promoter involved in meiotic cell cycle

pgaudet:
OK for SGD as well: 

> Hi Pascale,

I took a look at those annotations and the publication and I agree that they should be rehoused. I moved them to child terms underneath GO:0010672 regulation of transcription from RNA polymerase II promoter involved in meiotic cell cycle as suggested. No objection here to obsolete that term.

Regards,
Edith
********************************************************************

On Mar 13, 2017, at 2:21 PM, sgd-helpdesk-bounces@lists.stanford.edu wrote:


From: Pascale Gaudet <pascale.gaudet@sib.swiss>
Subject: obsoletion: GO:1900401 regulation of meiosis by regulation of transcription from RNA polymerase II promoter
Date: March 13, 2017 at 8:24:03 AM PDT
To: <sgd-helpdesk@mailman.stanford.edu>


Dear SGD curators, 

I am working on a issue brought up by Val, in which she requests the obsoletion of GO:1900401 regulation of meiosis by regulation of transcription from RNA polymerase II promoter. The reason is that "During previous changes this was re-defined as regulation of "meiotic cell division”, but the paper which is used to define the term is not for the "regulation of meiotic division", it is "regulation of meiotic cell cycle"
I'm not sure that meiotic division is regulated transcriptionally...(it's causally upstream). 

SGD has 2 experimental annotations

these annotations should be rehoused under:
GO:0010672 regulation of transcription from RNA polymerase II promoter involved in meiotic cell cycle”. 


If it’s OK with you I’ll go ahead with the obsoletion. 

Thanks, 
Pascale

pgaudet:
Hello, I am ready to do this. Looks like @mah11 created the term 
https://github.com/geneontology/go-ontology/issues/9370
[...]

PR #13165, opened 2017-03-14
Issue 13137

pgaudet:


pgaudet:
Fixes #13137

ukemi (changes requested):
In your first commit, it looks like you used has_part rel tions in the equivalence axioms for the second and third terms (GO:0010673 and GO:0010674). I think those should be part_of.

ukemi (approved):

PR #13186, opened 2017-03-17
new BP & MF terms for intermembrane lipid transfer

krchristie:


ukemi (commented):
@krchristie 
These terms all look fine based on our call yesterday. Would it be appropriate to create the part_of links between the various functions and the corresponding processes?

PR #13187, opened 2017-03-17
three new peptide hormone binding terms

krchristie:
@ukemi - Check the logical defs. After I did the equivalence axioms for two of these terms, which generate the correct inference:
SubClass of 'peptide hormone binding'

I looked at the logical defs of the other two peptide hormones that have equivalence axioms, 'insulin binding' and 'corticotropin-releasing hormone binding' and these two are defined with respect to 'binding' and then SubClass of 'peptide hormone binding' is asserted directly, instead of the way I did it where 'peptide hormone binding' is used in the equivalence axiom. The way I did it seems more concise, so I was wondering if there's something I'm missing about why these other two were logically defined with respect only to 'binding'


ukemi (approved):

ukemi:
If somatostatin and glucagon are already asserted to be peptide hormones in ChEBI, then all you need is binding as the genus. The inferences should still be made correctly to be children of peptide hormone binding. You can give it a try and change if you want. Sorry this comment was late. In general it is good practice to use the most general genus possible. It gives you the most powerful reasoning.

krchristie:
Yes, I can just do the equivalence axiom using 'binding', and the SubClass of 'peptide hormone binding' is still inferred by the reasoner.

Should I also remove the manual assertions **SubClass of 'peptide hormone binding'** for the two previous terms, since it seems redundant?
- 'insulin binding' 
- 'corticotropin-releasing hormone binding'

krchristie:
Actually, it doesn't seem to work to remove the direct assertion of SubClass of 'peptide hormone binding' for 'insulin binding', but if I remove this for the term 'corticotropin-releasing hormone binding', the SubClass of 'peptide hormone binding' relationship is inferred.

dosumis:
@krchristie check logical def insulin binding.  If that points to a Pro term then the inference should work.

krchristie:
It looks to me like the 'insulin' referred to in the logical def of 'insulin binding' is a PRO term.

Nevertheless, I've tried this several times and when I delete the directly asserted SubClass of 'peptide hormone binding' relationship from 'insulin binding', the relationship between 'insulin binding' and 'peptide hormone binding' just disappears and does not show up in yellow as an inferred relationship.

krchristie:
I fixed my terms and also removed an explicit subclass relationship for the previously existing term 'corticotropin-releasing hormone binding' since the reasoner inferred that subclass.

I did not remove the asserted subclass for 'insulin binding' because when I tried this (multiple times), the reasoner did not assert the subclass, so I just left it the way it was. I'm not going to look into this further.

dosumis:
> On Mar 21, 2017, at 6:49 PM, Karen R Christie <notifications@github.com> wrote:
> 
> I did not remove the asserted subclass for 'insulin binding' because when I tried this (multiple times), the reasoner did not assert the subclass, so I just left it the way it was. I'm not going to look into this further.

Fair enough.  For the record, the missing piece here is an assertion that insulin is a peptide hormone.  ‘coriticotropin releasing hormone’ has this classification.  It comes from ChEBI.  So, there is a general issue to fix here to do with reconcling ChEBI roles with proteins/peptides classified by PRO.

CC @cmungall

ukemi:
We need to import insulin from ChEBI, but I don't the the import request file any more.

krchristie:
When I looked at insulin in ChEBI, I thought it was classified as a peptide hormone there.

ukemi:
It is but it is not available to us because it is not imported, so you can't make a logical definition with it. We used to have a separate file to request terms that we needed. I don't see that file now. @cmungall

PR #13198, opened 2017-03-21
Fixed Subclass and definition, Fixes #13191

### LINKED ISSUE #13191: OTR: RISC-loading complex parentage
Hello,

GO:0070578 RISC-loading complex is described in the GO definition as "composed of the endonuclease Dicer (DICER1) and TRBP (TARBP2) in association with the Argonaute protein Ago2 (EIF2C2/AGO2). Within the complex, Dicer and TRBP are required to process precursor miRNAs (pre-miRNAs) to mature miRNAs and then load them onto Ago2. Ago2 bound to the mature miRNA constitutes the minimal RISC and may subsequently dissociate from Dicer and TRBP. This complex has endoribonuclease activity." 
Also, the papers supporting this term (PMID:18178619 PMID:19820710) only describe three protein members of this complex. Therefore I don't think this term should be is_a  GO:0030529 intracellular ribonucleoprotein complex, since the miRNA is not considered part of the complex, only a substrate of the endoribonuclease activity the complex has.

Could you please review this? If you agree, the first sentence of the definition should also be altered as it now states "A trimeric ribonucleoprotein complex that and is required for the formation of a mature RNA induced silencing complex (RISC).", which doesn't match the description that comes after.

Thanks,
Rachael.

pgaudet:


ukemi (changes requested):
You need to assert an is_a parent if none is inferred. In this case none will be inferred because there is no equivalence statement. Sorry if I confused you this morning about this.

pgaudet:
Thanks David ! I think I was looking at the inferred parent. 

Question: Can I add a logical definition containing a process? I was thinking of 
'macromolecular complex' and  ('capable of' some 'endoribonuclease activity') and ('part of' some 'small RNA loading onto RISC') 

Otherwise, the logical definition of the parent (endoribonuclease complex) is not unique to that term: 'macromolecular complex'  and ('capable of' some 'endoribonuclease activity')

Thanks, Pascale

ukemi:
For processes, you can use the capable_of_part_of relation. It means that the complex is capable of a function that is always part of a process.

ukemi (approved):
Looks good!

PR #13221, opened 2017-03-24
Issue 13220

pgaudet:


ukemi (commented):
I still want to look at this one on the call because although it looks correct, I'm not sure why the disjoint file was modified.

cmungall:
I'm not sure why this would have been modified. However, the change is harmless as it only pertains to OWL declaration axioms (too much detail: these are optional in the RDF/XML format that the disjoints file is saved on). It's perfectly safe to merge this request.

But it's always a good idea to do a `git status` before committing. If there are other files other than go-edit.owl changed, then do a:

```
git commit -m 'blah blah' go-edit.owl
```

rather than the full directory with `.`, this will ensure only the changes to the edit file go through to the PR. Then do a `git diff` and send the results to the go-ontology list for us to check if something weird is going on.

ukemi:
Thanks @cmungall 
Note that to use the command above, you need to navigate to the correct sub-directory. Otherwise you need to specify the path. I wonder if we should just make this an SOP. In the documentation, does it make sense to have editors navigate to the ontology directory and do the pull while on the master branch, then only deal with go-edit? Right now it is a bit misleading. We say to navigate to src and then the rest is from go-ontology. Note that you can also tell whether more than one file has changed in the diff.

pgaudet:
One question: it seems OK that the x-disjoint.owl file changed, since the old ID is referred to in there. Otherwise, how does it get updated?

cmungall:
On 24 Mar 2017, at 8:49, David Hill wrote:

> Thanks @cmungall
> Note that to use the command above, you need to navigate to the 
> correct sub-directory.

Ah yes, I forgot to note that. I always work from src/ontology

> Otherwise you need to specify the path. I wonder if we should just 
> make this an SOP. In the documentation, does it make sense to have 
> editors navigate to the ontology directory and do the pull while on 
> the master branch, then only deal with go-edit?

I think it may be easier if we just have one way of doing things, so 
yes.

cmungall:
On 24 Mar 2017, at 8:53, pgaudet wrote:

> One question: it seems OK that the x-disjoint.owl file changed, since 
> the old ID is referred to in there. Otherwise, how does it get 
> updated?

Good question. The main way it would get changed if someone explicitly 
adds a new disjointness axiom (we don't have an SOP for this - this 
takes a bit of care as the axiom should be added to the imported 
ontology rather than the parent. The easiest way is to highlight the 
axiom and choose  'move to ontology' in Protege).

The other possible change which is undocumented is if the disjointness 
axiom refers to a class that is being obsoleted. We delete all logical 
axioms for an obsolete class (or the plugin does this, when it works). 
This should result in an axiom deletion in x-disjoints.

I don't think there are any other scenarios where the logical content of 
x-disjoints changes.

ukemi:
> The other possible change which is undocumented is if the disjointness axiom refers to a class that is being obsoleted. We delete all logical axioms for an obsolete class (or the plugin does this, when it works). This should result in an axiom deletion in x-disjoints.

Maybe this explains it. In this case it followed a merge, which in the new world actually gets treated as an obsoletion of the losing term, right?

PR #13242, opened 2017-03-28
Updating logical def for GO:1990771 Fixes #13211

### LINKED ISSUE #13211: Incorrect parent 'clathrin-dependent extracellular exosome endocytosis'
Hello,

The term 'clathrin-dependent extracellular exosome endocytosis' has parent 'extracellular transport'. This seems wrong; I suggest moving it under 'endocytosis'. 

This terms is only used by RGD, @slaulederkind is this OK with you ? 

Thanks, Pascale

vanaukenk:
I've updated the logical definition for GO:1990771 from:

'clathrin-dependent endocytosis' and ('transports of maintains localization of' some 'extracellular exosome')

to:

'clathrin-dependent endocytosis' and ('has participant' some 'extracellular exosome')

If we need to re-visit the logical def at some point, I can do that, but this fixes the immediate issue of incorrect parentage for now.

@ukemi  - if you just could take a quick look to make sure everything looks okay, that'd be great.  Thx.


ukemi (commented):
I'm still not convinced that this logical definition is N and S. I would be tempted to remove the logical def altogether and just make the relationships. @dosumis , can you comment? If I'm not mistaken, I think you had done some work with patterns for these terms.

cmungall:
Shouldn't it have as a participant:

     GO:0045334 ! clathrin-coated endocytic vesicle

Although has_participant is a general relation, I feel this wouldn't be 
over-inclusive (though my tendencies are a bit looser than @dosumis')

However, I think in cases like this where we can define a process by 
existing named subprocesses, we should do that.

e.g:

     id: GO:0099042
     name: nucleation of clathrin-coated pit

I would use has_part in the equivalence axiom, and add additional 
conditions with subclass, e.g. has_part NoCCP

Now we're in Protege there's nothing to stop you putting a class-level 
Noctua model in:

     endocytosis and starts-with some (NoCCP and provides-input-for some 
(clathrin-coated pit assembly and ...

dosumis:
> Now we're in Protege there's nothing to stop you putting a class-level
Noctua model in:

>      endocytosis and starts-with some (NoCCP and provides-input-for some
(clathrin-coated pit assembly and ...

doesn't play nicely with OBO master.

cmungall:
On 28 Mar 2017, at 9:01, David Osumi-Sutherland wrote:

> doesn't play nicely with OBO master.

Yes, would have to live in an import

dosumis:
has_participant is not wrong and I'm confident the resulting def is necessary and sufficient.

BUT: IIRC this process preserves the exosomes as vesicles inside a larger endocytic vesicle  (exosomes can certainly be produced by the exocytosis of a large vesicle containing them). If that is the case, how should curators annotated these vesicles  after they are internalized (or before they are released).  'extracellular exosome' would clearly be wrong as it would imply an extracellular location.

ukemi:
If annotators still refer to them as 'extracellular exosome', then we would need to remove the part_of relation between it and 'extracellular space'. Based on the term as it stands, I suspect the fact that is is part of the extracellular space is important. But it seems weird to have to create additional terms to describe them based on before exocytosis and after endocytosis as well. Remove the part_of in the ontology and only have it in the instance-based models?

vanaukenk:
Hi,

From a review article (PMID:27875100), here is a statement I have found on intraluminal vesicles and exosomes:

- By contrast, exosomes (below 150 nm diameter) originate from an endosomal compartment, the MultiVesicular Body (MVB), through a variety of possible molecular mechanisms (reviewed
in ref.17 and discussed thereafter). MVBs contain small intraluminal vesicles (ILVs), which, upon secretion outside of cells, are named exosomes.

This would suggest that when curators want to annotate the intracellular vesicle, they should use the existing GO term, 'GO:0097487 multivesicular body, internal vesicle', and when they want to annotate the extracellular vesicle, they should use 'GO:0070062 extracellular exosome' (which I am beginning to think should simply be called 'exosome', an existing exact synonym).

For GO:0097487, we could perhaps also add exact synonyms of 'multivesicular body, intraluminal vesicle' and 'ILV'.

We could then preserve the extracellular parentage for exosome, which I think is important for the biology, but still modify the existing logical definition as originally suggested so that we don't get incorrect parentage for 'GO:1990771 clathrin-dependent extracellular exosome endocytosis'.

Wrt the logical definitions, I would like more feedback and discussion on the editors calls about how we want to approach these.  What is the knowledge threshold for determining if something is N and S?  If we don't know of an exception, do we add the logical def and just remove it if and when we do come across an exception?   

Also, the Noctua-ish logical defs are very appealing from the standpoint of knowledge modeling, but what are our expectations about if/how these should be created?  And what does it mean to say something doesn't 'play nice with OBO master and would have to live in an import'?

PR #13279, opened 2017-03-31
Issue 13263

vanaukenk:
Obsoletions for tRNA methylation involved in response to nitrogen starvation and regulation term.  New terms for tRNA methylation, regulation and tRNA C5-cytosine methylation, regulation.

@ukemi  - for some reason the Reviewers widget is getting hung up on 'Loading suggestions...' so I can't flag this for review but if you could take a look to see if there's anything that needs fixing, that'd be much appreciated.  Thx.


ukemi:
@vanaukenk  This all looks good to me, (disclaimer that I am notorious for missing typos) but as I was reading the diff I wondered if we remembered to include adding replaced_by and consider tags to obsolete terms in the SOPs. If not, we should.

vanaukenk:
Thanks @ukemi   I've added info about replaced_by and consider to the SOP Google doc.

Issue #13308, opened 2017-04-06
Makefile should detect equivalencies between named classes in GO


cmungall:

dougli1sqrd:
https://github.com/geneontology/go-ontology/pull/15060

PR #13310, opened 2017-04-06
WIP: Robot flag to disallow equivs

cmungall:
Note: too senstive, need to allow CHEBI to PR
see #13308 
DO NOT MERGE YET


cmungall:
```
ERROR Equivalency: <http://purl.obolibrary.org/obo/SO_0000252> == [<http://purl.obolibrary.org/obo/CHEBI_18111>] // null == [null]
ERROR Equivalency: <http://purl.obolibrary.org/obo/SO_0000234> == [<http://purl.obolibrary.org/obo/CHEBI_33699>] // null == [null]
ERROR Equivalency: <http://purl.obolibrary.org/obo/CHEBI_18111> == [<http://purl.obolibrary.org/obo/SO_0000252>] // null == [null]
ERROR Equivalency: <http://purl.obolibrary.org/obo/CHEBI_36080> == [<http://purl.obolibrary.org/obo/PR_000000001>] // null == [null]
ERROR Equivalency: <http://purl.obolibrary.org/obo/CHEBI_33699> == [<http://purl.obolibrary.org/obo/SO_0000234>] // null == [null]
ERROR Equivalency: <http://purl.obolibrary.org/obo/PR_000000001> == [<http://purl.obolibrary.org/obo/CHEBI_36080>] // null == [null]
ERROR Equivalency: <http://purl.obolibrary.org/obo/CHEBI_17843> == [<http://purl.obolibrary.org/obo/SO_0000253>] // null == [null]
```

Need to extend robot, either whitelist of allowed equivalence pairs OR allow inter-ontology (with prefix detection hack) OR simply add asserted equivalences between named classes to whitelist

dosumis:
These inter-ontology links doing important work so 👍  to extending Robot.

cmungall:
@dosumis - comments welcome on https://github.com/ontodev/robot/issues/161

dougli1sqrd:
Just so I understand what's happening, when we told robot to not allow equivalent classes, these then showed up in @cmungall's log output above, but those equiv classes are okay? Thus the build fails when it shouldn't?

dosumis:
> Just so I understand what's happening, when we told robot to not allow equivalent classes, these then showed up in @cmungall's log output above, but those equiv classes are okay? Thus the build fails when it shouldn't?

Equivalence between named classes is never OK within a single ontology.  We still want a warning for that.  For some projects at least, it is required between terms from different imported ontologies.  

One way to cope with this is to allow a mechanism by which a white-list of known equivalent classes could be specified.  But this would be hard to maintain.  Better would be an option that allows equivalence between terms from different source ontologies e.g.

If ontology X imports ontologies Y and Z

* OK:
   * class A (in X) EquivalentTo: Class B (in Y)
   * class B (in Y) EquivalentTo: Class C (in Z)
* Not OK:
   * Class A (in X) EquivalentTo: Class D (in X)
   * Class E (in Y) EquivalentTo: Class F (in Y)

dougli1sqrd:
Thanks @dosumis for the explanation! So in the above log output, those equivalences are actually okay because they're across different ontologies.

cmungall:
As soon as @dougli1sqrd's PR makes it into a robot release we can update the robot arguments in this makefile

cmungall:
Currently failing as these are not asserted:

```
ERROR Equivalence: <http://purl.obolibrary.org/obo/CHEBI_18111> == <http://purl.obolibrary.org/obo/SO_0000252>
ERROR Equivalence: <http://purl.obolibrary.org/obo/CHEBI_36080> == <http://purl.obolibrary.org/obo/PR_000000001>
ERROR Equivalence: <http://purl.obolibrary.org/obo/SO_0000253> == <http://purl.obolibrary.org/obo/CHEBI_17843>
ERROR Equivalence: <http://purl.obolibrary.org/obo/CHEBI_33699> == <http://purl.obolibrary.org/obo/SO_0000234>
```

dosumis:
They are asserted in GO_bridge.owl

But there seems to be something weird going on with the imports. The axioms from go-bridge aren't showing up in Protege (5.2) after loading go-edit.  Don't understand why as the import statement is present and the file is loaded for the session.

cmungall:
I'll close this and make a new cleaner PR where we simply add

`--equivalent-classes-allowed asserted-only`

to the reason command

cmungall:
Ah looks like @balhoff already did this in 35243166d awesome!

balhoff:
It was some time before, I think you must have added it.

PR #13319, opened 2017-04-06
Issue 13218 2

pgaudet:
I also fixed the two siblings 'G1/S transition of mitotic cell cycle' and 'metaphase/anaphase transition of mitotic cell cycle' to have comparable subclasses and logical definitions.


ukemi:
In the first stanza, doesn't this:
is_a: GO:0044772 ! mitotic cell cycle phase transition
get inferred?

PR #13370, opened 2017-04-18
fixed term name and definitions Fixes #13354

### LINKED ISSUE #13354: rename	GO:0031074	nucleocytoplasmic shuttling complex
GO:0031074	nucleocytoplasmic shuttling complex
Any complex that acts to move proteins or RNAs into or out of the nucleus through nuclear pores.

"nucleocytoplasmic shuttling protein" is normally used to describe proteins which are 'shuttled', not the transport complexes. I have community curators using for "shuttled proteins".

 - [x] Could this grouping term be renamed  "nucleocytoplasmic transport complex"? 
- [x] Moves in def should change to transports"
- [ ]  flagged term not for direct annotation"

Some of the  descendent complex terms have "shuttled" in the definition. This should probably also be "transports".  Usually the transport complex is itself "shuttled" but the transport complex is not "shuttling". "Shuttled" implies the complex moves the cargo in both directions, but here is used to describe unidirectional cargo.

pgaudet:


cmungall:
```
--------------------------------------------------------------------------------
Parser: org.semanticweb.owlapi.oboformat.OBOFormatOWLAPIParser@2e54db99
    Stack trace:
LINENO: 199679 - Unexpected tag subsetdef in term frame.
LINE: subsetdef: gocheck_do_not_manually_annotate        o
```

should be inSubset OWL tag

PR #13281, opened 2017-04-01
Removing redundant subclass assertions. 

cmungall:
Fixes #13275

PR #13397, opened 2017-04-21
Removing redundant subclass assertions. 

### LINKED ISSUE #13275: removal of asserted regulates relationships
In discussion of the issue about the appropriate pattern for making regulation terms that was brought up in issue https://github.com/geneontology/go-ontology/issues/13252

we agreed that the directly asserted SubClass Of relationships of the form:
**regulates** some **'Kit signaling pathway'**

should be removed computationally for the many regulation terms where currently exist.

cmungall:
Fixes #13275


cmungall:
This replaces #13281

dosumis:
Looks good to me.  Would be good to do something similar to remove directly redundant 'is_a' assertions.

Issue #13472, opened 2017-05-04
Put up prototype of TG replacement before Corvalis meeting


cmungall:
We will use table-edit/taboggan as TG replacement. Essentially same operation as adding a row.

 * [x] move repo containing DOSDP yamls #13273 
 * [ ] configure to use the relevant autocomplete: GO, Uberon, CL, PO, OBA, PATO, CHEBI
 * [ ] seed a csv from the ontology plus a given DP using @balhoff's https://github.com/balhoff/dosdp-scala, place in repo
 * [ ] configure a table-edit page (preferably with a 'badge' ([see](https://github.com/INCATools/table-editor/issues/6) )
 * [ ] demonstrate a cycle of adding a new row, saving, run reasoner

jmcmurry:
cc: @doctorbud, @dougli1sqrd

ValWood:
Yay... we miss TG....

dougli1sqrd:
https://github.com/geneontology/go-ontology/pull/13486 for the first point

cmungall:
For @DoctorBud's demo, we will use this tsv that we re-engineered from the OWL using @balhoff's dosdp-tools: [src/design_patterns/validated_patterns/activator_activity.tsv](https://github.com/geneontology/go-ontology/blob/tg-demo/src/design_patterns/validated_patterns/activator_activity.tsv_

(this is on the tg-demo branch in Pr #13566)

The corresponding yaml is [src/design_patterns/validated_patterns/activator_activity.yaml]
(https://github.com/geneontology/go-ontology/blob/tg-demo/src/design_patterns/validated_patterns/activator_activity.yaml)

(also on master)

dosumis:
I already built that tsv. See pattern implementation tsv dir.

DoctorBud:
Closing this ticket because the goal has been achieved, the demo is over, and the time has passed. Even though there is rich reference material here, the ticket should not be on the active list.

dosumis:
#13692 New umbrella ticket covering steps needed to get the TG replacement up and running.

Issue #13273, opened 2017-03-31
Move design patterns into main go-ontology repository


pgaudet:

cmungall:
is this about the location of the yaml design patterns?

dosumis:
Do you mean design pattern docs?

cmungall:
I modified the title; yes, this is about moving https://github.com/geneontology/design_patterns into this repo.

It would make it easier for editors to find. I think it also makes sense to have a single tracker since the DPs are so integrally part of what we do on go-ontology. Also for CI checks like travis etc it's slightly simpler to have everything in one place.

Would moving be OK with you @dosumis? I suggest src/patterns. We could use the GH issue mover to move the existing tickets.

After that we can set up some additional jenkins jobs. For example, here is a job we use for the pheno ontologies: https://build.berkeleybop.org/job/query-pheno-patterns/lastStableBuild/ - this shows all terms that match each pattern

dosumis:
Makes sense for tickets to be integrated. 

But the advantage of a separate repo is that we can hook it up to a simple, separate travis job (with badge and emails warnings) that checks and reports when editors get them wrong and updates them when lable change. Simple checks are running now & I have some more in the works (although may be superceded by Jim’s Scala code). 

I’d like to have a standard recepe for pattern repos, just as you have for ontology repos - with travis pulling some standard checking code.

Ot - can see a way to do this cleanly within the ontology repo?

cmungall:
Looks like I lost my comments here.

We of course still want to run the travis checks over the DPs. Having in the same repo will also allow us to check for things such as modifying a DP having implications for the ontology

cmungall:
Decided on http://wiki.geneontology.org/index.php/Ontology_meeting_2017-04-21

We will first check that we will still get fast travis feedback on any edits to the DPs, and if this is OK, move to src/patterns

dougli1sqrd:
https://github.com/geneontology/go-ontology/pull/13486

dosumis:
So - the files have been moved, but no sign of integrating into travis yet.  It looks like simply specifying another travis step with is not a good idea due to lack of support for multiple languages in one .travis.yml, so this should probably be incorporated into the makefile.  Also the badges are still linking back to the travis job on the old repo...  

I'm actively editing patterns and would like to have checking in place while I do so.

cmungall:
Giving up for now, @dougli1sqrd can you take a look at my WIP PR above?

cmungall:
[...]

PR #13486, opened 2017-05-10
moving contents of geneontology/design_patterns to go-ontology/src maintaining history

dougli1sqrd:
In regards to https://github.com/geneontology/go-ontology/issues/13273 and https://github.com/geneontology/go-ontology/issues/13472


cmungall (approved):
Note the .travis.yml will have no effect in the subdir, but we can merge it into the main one later

Issue #12946, opened 2017-01-20
set up BFO import job


dosumis:
extensions/go_bridge.owl contains BFO classes. We need a standard BFO import file pulling relevant classes and axioms (APs, disjointness) using these classes a seed.

dosumis:
Now we have Arachne running in Noctua, this is causing distracting high level inferences to show up in Noctua as well as in inferences.  Would be good to get this fixed before Corvallis meeting.

PR #13523, opened 2017-05-19
RO import modules now includes BFO classes

cmungall:
See #12946


dosumis:
Why via RO rather than making a seed using BFO classes references in go_bridge.owl?

cmungall:
I'm going ahead and merging, we can discuss refinements to the pipeline later

PR #13547, opened 2017-05-26
WIP DO NOT MERGE: DP travis fixes, #13273

cmungall:
See #13273


dosumis:
Looks good as far as I can tell.  BTW - better script in utils (simple_pattern_checker.py).  (Need to centralise all this so Python checking scripts pulled directly from DOSDP github).

PR #13566, opened 2017-05-30
WIP: Fixes to YAML DPs, and generated TSVs

cmungall:
See #13472


dosumis:
Looks ok, but limited coverage. More complete tsv files for regulator activities here: https://github.com/geneontology/go-ontology/tree/master/src/pattern_implementation_tsv
Tested activator activity tsv with dosdp-scala.

cmungall:
Ah didn't see that directory! So how did you generate these?

dosumis:
> On Jun 1, 2017, at 1:41 AM, Chris Mungall <notifications@github.com> wrote:
> 
> Ah didn't see that directory! So how did you generate these?

Python munging from obo graphs json.

Would be easier with OWL queries though, so likely to do scala (+amn repl?) in future.

PR #13587, opened 2017-06-03
Virus part2

pgaudet:


dosumis:
This new logical def for imoort into Cajal body looks wrong:

 +intersection_of: GO:0006810 ! transport
 +intersection_of: has_target_end_location GO:0005634 ! nucleus
 +intersection_of: has_target_start_location GO:0005737 ! cytoplasm
 +intersection_of: imports SO:0000274 ! snRNA
 +intersection_of: results_in_transport_across GO:0015030 ! Cajal body

End location should be Cajal body.  No need for results in transport across.  Start location maybe nucleus?

PR #13625, opened 2017-06-10
Go synapse 140

dosumis:
Important changes to review are in https://github.com/geneontology/go-ontology/pull/13625/commits/5e4438d1f4ba5aa61fa040de8904fa8664c40db6

geneontology/synapse#139 and geneontology/synapse#140 may need further discussion: What are they intended to be used for besides actins?
Should the synaptic nature of these be recorded in extensions?


dosumis:
Merging following assurance on original ticket about cope of usage.

Issue #13669, opened 2017-06-16
Ensure pipeline checks for spaces in attribution IDs


cmungall:
Looks like a bad definition xref escaped: https://github.com/geneontology/helpdesk/issues/28

Our current ID checks rely on a perl check of the obo which is convenient but limited. If we augment these with SPARQL checks we can be more complete.

As it happens I implemented this for HPO: https://github.com/obophenotype/human-phenotype-ontology/pull/2376

@dougli1sqrd should we use SPARTA for running the checks in the GO ontology or is it more convenient just to use the robot `validate` option you implemented? Recall this is on the edit version of the ontology prior to loading into any triplestore

cmungall:
We already have some sparql here: https://github.com/geneontology/go-ontology/tree/master/src/sparql

And a section of the Makefile for reports:

https://github.com/geneontology/go-ontology/blob/master/src/ontology/Makefile#L324-L340

Seems most straightforward to just go ahead and switch this from using jena-arq to the more portable robot. But we should think about how we might share many of the checks across ontologies, cc @drseb @jamesaoverton

jamesaoverton:
I guess it's a choice between a new repo for OBO checks or just putting them in the ROBOT repo. I can see pros and cons to both.

PR #13697, opened 2017-06-21
SPARQL violation checks

cmungall:
#13669


cmungall:
Argh, looks like travis is killing the job as it takes too long to do the SPARQL queries?

cmungall:
Hmm, takes only 23s on my laptop, that can't be the reason..

PR #13795, opened 2017-07-03
DO NOT MERGE YET: Adding SPARQL check to reference to obsolete classes

### LINKED ISSUE #13793: Add SPARQL check to ensure that no classes reference obsolete classes
To prevent this: https://github.com/geneontology/go-ontology/issues/13685#issuecomment-312710378

cmungall:
Fixes #13793


cmungall:
This seems to be taking too long

cmungall:
@dougli1sqrd - now we have an explicit verify command we don't need the perl hack as well?

dougli1sqrd:
Yeah that's right: https://github.com/geneontology/go-ontology/pull/14097/files#diff-22e910b7915a0c24f526851b8507813dR376 
But looking at the pull request here, it looks like that merge never happened, or was re-written over?

yy20716:
* ok, I rewrote obsolete-reference-violation.sparql to reduce its execution time as requested. Unfortunately, sometimes the order of the triple patterns in a query matters in Jena.

PR #13807, opened 2017-07-05
Various syngo

dosumis:
Minor edits and term requests from syngo.  See individual commits for ticket numbers/links.


dosumis:
Diffs look good and changes are simple enough not to need third party review, so merging.

Issue #13824, opened 2017-07-06
Rename 'neurological system process' to 'nervous system process'


pgaudet:
This was brought up in #13806

@pgaudet: 
> I wonder, should we refer to 'neurological system', since this is how it's described in GO (AFAIK). (I always wondered why we didn't have 'nervous system process', but this may be a whole different issue).

@dosumis 
> I think nervous system is fine. People are more likely to understand it. We could add a formal clause
linking behavior to Uberon nervous system:

SubClassOf 'has participant' some 'nervous system'

'neurological system process' looks like it might be broader. Should be aligned with Uberon and ideally get some kind of logical def that improves the weirdly spotty classification under it. But that should be another ticket.

dosumis:
My point was the 'nervous system' was fine for the definition of behavior, rather than the broader 'neurological system' that this term refers to.  It might be worth keeping 'neurological system process' if the intent is broader and others think it useful.  Changing to 'nervous system process' would probably mean find a home for some of its children.   Should check whether Uberon has 'neurological system'

pgaudet:
Hi @dosumis 

I don't mind one way or the other. You can just close this issue if this is not appropriate.

pgaudet:
Hi @dosumis 
Uberon has 'neurological system' as an exact synonym for 'nervous system'. 
http://www.ontobee.org/ontology/UBERON?iri=http://purl.obolibrary.org/obo/UBERON_0001016

... let me know what you would like me to do. 

Thanks, Pascale

pgaudet:
Hi @dosumis 

Just as I did the merge request I saw what you were saying. Is there any way to revert it ? 

I do think it’s a problem that Uberon and GO do not use the same meaning for neurological/nervous system processes. It seems to me Uberon should be the authority on this. So if we assume that neurological system is an exact synonym of nervous system as stated in Uberon, I can finish the implementation of this in GO (ie fix the regulation classes and the definitions that use ‘neurological process’). 

What do you think ?

pgaudet:
I cannot find anywhere a definition of neurological system versus nervous system. "Neurological" seems to be more used in the medical field, for example 'Neurological disorders', which are described as 'A neurological disorder is any disorder of the nervous system' (https://en.wikipedia.org/wiki/Neurological_disorder). I am closing this one.

pgaudet:
From @judyblake 

From one of our resident neurologists here at JAX

Neurology:  study of diseases of nervous system.
Nervous System:  nervous system

[...]

PR #13825, opened 2017-07-06
changed label, added synonym and logical definion as I understood #13824

pgaudet:
@dosumis please review - merge or close without merging. 

Thanks :)


dosumis:
If you were to make this change, you should also review the children to see if they fit with being nervous system processes - given Uberon's definition of a nervous system.

It would be better if we could come up with a logical def that allowed us to leverage Uberon for classification under here, but that may be hard.

pgaudet:
I am copying my comment from related issue #13824 

Uberon has 'neurological system' as an exact synonym for 'nervous system'. 
http://www.ontobee.org/ontology/UBERON?iri=http://purl.obolibrary.org/obo/UBERON_0001016

 @dosumis  let me know what you would like me to do. 

Thanks, Pascale

PR #13834, opened 2017-07-06
Adding legorel, was never fully ported from svn

cmungall:
Do we need this long term? See https://github.com/geneontology/minerva/issues/121


cmungall:
Will not be visible till tomorrow - can create a purl for now

PR #13850, opened 2017-07-07
Renamed generic terms and relaced macroautophagy term names with curr…

ukemi:
…ent terms. Created selective autophagy.


ukemi:
DO NOT MERGE YET. To be discussed on a call on Monday.

PR #13899, opened 2017-07-13
Added a do not annotate class for general autophagy that is not catab…

### LINKED ISSUE #13886: Cvt pathway should be an autophagy
After community feedback, we have decided to make the Cvt pathway a type of autophagy. Since currently autophagy is defined as being a catabolic pathway, we will change the name of the current term to catabolic autophagy and create a new autophagy grouping class. Catabolic autophagy and the Cvt pathway will then become children of the new class.

ukemi:
…olic and allows for grouping of the Cvt pathway. Fixes #13886


ukemi:
DO NOT MERGE

ukemi:
Merged after call.

Issue #13916, opened 2017-07-17
NTR: membrane curvature sensing


BarbaraCzub:
Figure 4 and Supplementary Figure 4 in PMID: 25898166 demonstrate that PICALM is involved in sensing of membrane curvature. 

In order to annotate this I would like to request a new term: 'membrane curvature sensing'.

_Suggested_ definition: "A process involving (1) the insertion (aka wedging) of the hydrophobic residue side chains of the amphipathic helix (AH) of a membrane curvature sensing (MCS) protein into lipids in a curved membrane and (2) the interactions between the hydrophilic residue side chains of the AH and membrane phospholipid headgroups. Membrane curvature sensing is a part of 'GO:0061024 membrane organization' and 'GO:0016050 vesicle organization'."
Refs: PMID: 21953452 and PMID: 25898166.

_Suggested_ synonyms:
_Exact synonym_: membrane amphipathic lipid packaging sensing (PMID: 25898166 and references within)
_Related synonym_: membrane curvature-sensitive protein-membrane binding (PMID: 21953452).
_Related synonym_: membrane curvature-depending protein-membrane binding
_Related synonym_: membrane curvature-sensitive protein-membrane interaction
_Related synonym_: membrane curvature-depending protein-membrane interaction

> GO:0061024 membrane organization
>> NTR membrane curvature sensing (part_of)

> GO:0016050 vesicle organization
>> NTR membrane curvature sensing (part_of)

Please use refs: GOC:aruk and GOC:bc.

Thanks,
Barbara

pgaudet:
@BarbaraCzub 
Shouldn't the term refer to the vesicle ? ie something like 'clathrin-coated vesicle membrane curvature sensing'?

BarbaraCzub:
Hi @pgaudet 

(re: 1) I am not convinced. There are membranes that are not vesicle membranes, which are certainly curved. The two papers, which I have referenced involved studies done on liposomes, so we should not create organelle-specific terms, based on them.

(re: 2) For 'is_a' parent I would be reluctant to choose anything more specific than 'GO:0051128 regulation of cellular component organization' before we can find references with relevant evidence. However, we could create a more specific new parent 'regulation of membrane organization'.

Thanks,
Barbara

ValWood:
I'm a bit unsure about the term request. what is the process of "membrane curvature sensing"
 If it is a single gene product that can monitor membrane curvature in some way it sounds more like a molecular function. 

Then you would annotate 
[...]

PR #13963, opened 2017-07-21
Issue 13916

pgaudet:


cmungall:
Sorry, trying to figure out what happened in this chain of commits. It looks like you accidentally checked in the diff markers (`<<<<<` etc) when resolving a commit. 

What I expect happened next is that these markers caused a sequence of terms to be silently omitted when loading in Protege. We would like the parser to throw an error here but unfortunately the obo parser is quite lax. 

This meant that you lost your GO:0140063, see https://github.com/geneontology/go-ontology/pull/13963/commits/48018a553b90097b758c35ea8375f7f923d06f51

I could do a bit more detective work, but either way this PR has to die and be redone, sorry. It looks like you did work for other terms than #13916 on this PR, this will also need to be redone.

For example, the first commit here: https://github.com/geneontology/go-ontology/pull/13963/commits/151f7dcceecee99ce09e76162cf07084bd44ad26

say that it fixes 13916, but this term `cardiac endothelial to mesenchymal transition` had nothing to do with that ticket

I could go through the commits and paste in the obo stanzas you made but will have to wait.

Remember

 - always do a `git diff` before committing
 - try and separate different tickets into different branches/PRs
 - be especially careful resolving conflicts on the command line
 - if things start to look weird, consider abandoning the branch and starting again

PR #14025, opened 2017-08-08
WIP: Mf refactoring

dosumis:
Upper level changes for MF refactoring.  Edits here only cover top level terms.  There are separate branched for ATPase-coupled activities, receptors and transcription factor activities.

CC @thomaspd @cmungall @ukemi @vanaukenk


dosumis:
I should clarify - this is mostly upper level changes.  There are a few more low level ones that should probably be in other branches.  For now, it would be great if you could review the new upper level terms and changes to new upper level existing defs & comment here.  **Please don't merge yet.**

pgaudet:
@thomaspd @pgaudet
Some feedback: 
1. GO:0060089  name: molecular transducer activity
Update definition to: A compound molecular function in which an effector function is controlled by one or more regulatory *functions*.  (*was: components).

PR #14033, opened 2017-08-08
added 18 terms and additional dbxrefs as per ticket 10873

hdrabkin:


cmungall:
Apologies for the opacity of the error report.

@dougli1sqrd  will implement this which is the proper way to check https://github.com/ontodev/robot/issues/177

we are using a hacky perl wrapper over the sparql query. It should be reporting the axiom that has the trailing space but it isn't. I suspect you may have done something like pasted in an odd character

cmungall:
> So that's what the orange means; can't tell just by looking at line

The line is highlghted orange because I provided a hyperlink to that line.

hdrabkin:
Ah, well removing the extra space solved the problem and it appears there is nothing else wrong. Thanks for your help!

Issue #14041, opened 2017-08-09
An identifier was deleted rather than obsoleted/merged: GO:0044711  single-organism biosynthetic process


cmungall:
this is gone: no record, no alt_id, no obsolete class!!

```
[Term]
id: GO:0044711
name: single-organism biosynthetic process
namespace: biological_process
def: "A biosynthetic process - chemical reactions and pathways resulting in the formation of substances - involving a single organism." [GOC:jl]
is_a: GO:0009058 ! biosynthetic process
is_a: GO:0044710 ! single-organism metabolic process
intersection_of: GO:0009058 ! biosynthetic process
intersection_of: bearer_of PATO:0002487 ! single organismal process quality
created_by: janelomax
creation_date: 2012-10-17T15:52:18Z
````

 * [ ] add check back to travis @cmungall
 * [ ] reinstate class as obsolete @pgaudet

cmungall:
This wasn't caught by travis; @kltm caught it with the new complete pipeline

https://build.geneontology.org/job/geneontology/job/pipeline/job/snapshot/65/changes

![image](https://user-images.githubusercontent.com/50745/29147524-6da2d6f0-7d1c-11e7-9d6f-29432543aacd.png)

pgaudet:
Not sure what happened. I was trying to merge all these terms with their parent, so it should have been an alt ID of biosynthetic process. Looks like I deleted the ID after the merge (because '  single-organism biosynthetic process' was still a synonym of 'biosythetic process', so I did do a merge. I removed the synonym. 

Thanks for catching that ! 

Pascale

pgaudet:
I will add back the synonym and the alternative ID.

PR #14042, opened 2017-08-09
Restoring deleted ID and adding back change-report.txt to travis test

cmungall:
#14041


cmungall:
This fails as expected:

```
xref: MetaCyc:RXN-8282  has metacyc 2.4.1.67-RXN?
2017-08-09 23:13:42,094 ERROR (CommandRunner:5316) The verification failed with the following errors.
2017-08-09 23:13:42,094 ERROR (CommandRunner:5323) Missing ID: GO:0044711
make: *** [change-report.txt] Error 255
```

it should pass when #14042 is addressed

PR #14049, opened 2017-08-11
fixed incorrect namespace for obsoleted term

hdrabkin:


hdrabkin:
This was seen during a GO-load failure at MGI: 
The GO vocab load failed last night because one of the terms in the input file had no namespace
defined for it. Here is the term:
 
[Term]
id: GO:0044711
name: obsolete single-organism biosynthetic process
def: "OBSOLETE. A biosynthetic process - chemical reactions and pathways resulting in the formation of substances - involving a single organism." [GOC:jl]

I looked and the namespace was "gene_ontology'; changed to 'biological_process'

PR #14096, opened 2017-08-17
Merged mediated_by into has_active_participant

dosumis:
OBO file opens in Protege (or OWLtools), but then will not save as OBO.  I've struggled with this in the past.   IIRC - I found some v.weird fix that involved adding multiple labels...

CC @cmungall


dosumis:
Branch is passing tests.  But roundtrip with owltools or Protege still fails.  @cmungall - please can you look into this?

pgaudet:
Out of date.

dosumis:
How was this resolved?

PR #14121, opened 2017-08-26
Synapse requests

dosumis:


dosumis:
@cmungall Fixed error on last commit that should have caused a check fail for named equivalent classes in GO, but didn't.  So, looks like that check may be broken.

PR #14177, opened 2017-09-10
Synapse edits again

dosumis:


dosumis:
Weird build fail: https://travis-ci.org/geneontology/go-ontology/builds/273835808?utm_source=github_status&utm_medium=notification coming from Robot not recognizing file type.  The file was saved from Protege (no hand edits).  Can't spot anything odd in diff, + head&tail of OBO file look fine.

Seeing the same thing locally with robot.

PR #14291, opened 2017-10-04
Move IntAct xrefs from def xrefs to xrefs

cmungall:
cc @ukemi


cmungall:
note: needs a roundtrip to avoid spurious tag ordering diff

PR #14295, opened 2017-10-05
Removed functional characteristic from part_of.

balhoff:


ukemi:
Sees like this is fine to merge. I don't see why an ontology editor would have added it manually.

balhoff:
Thanks @ukemi. I don't have permissions, so please merge.

PR #14318, opened 2017-10-09
Tf refactoring

dosumis:
Still working on this.  Making pull request so that others can look/comment.

TODO:

 - [ ] split out name changes (from spreadsheet)
 - [ ] Implement text definition changes (from spreadsheet)
 - [ ] Finish formalization of 'protein binding' branch'


dosumis:
> Please do not merge this branch into the master branch. Pascale and I are working on a refactoring and will make sure as many of these changes as possible are incorporated into it.

I won't.  But will do some more work finishing up what I've done and documenting.  Then it's up to you.  Ditto the signalling.  For that I have an unpushed branch with has_ligand relationships.  Logical wiring means these logical defs can be used for inference even when has_ligand not directly used. 

Some thought has gone into making coherent patterns for all of these.  It would be a shame if that work was lost.  I'd really like to write some of it up.

pgaudet:
We will not implement this, since it's not completely consistent with the changes made by @thomaspd and I for the MF refactoring. 

However we'll need to create logical definitions I'll create a ticket for that. 

Pascale

PR #14408, opened 2017-10-23
created new term, fixed some logical definitions Fixes #14391

### LINKED ISSUE #14391: ntr: cellular detoxification of fluoride
_From @ValWood on October 20, 2017 9:40_

standard def

Any process carried out at the cellular level that reduces or removes the toxicity of a flouride. These may include transport of flouride away from sensitive areas and to compartments or complexes whose purpose is sequestration of the toxic substance.

does this also include "export"? this isn't clear from the standard def?

"Therefore, fluoride is toxic for S. pombe even at low concentrations, and specific channels that export fluoride ions are required for normal growth."

_Copied from original issue: geneontology/go-annotation#1665_

pgaudet:


cmungall (changes requested):
Wait, why are we changing the DP for GO:0140041 but leaving the others unchanged?

__UPDATE__

I'm not sure of our current DP for cellular-X. It seems we just define as X-and-occurs-in cell, but we don't have a yaml for this in https://github.com/geneontology/go-ontology/tree/master/src/design_patterns

If this is indeed our DP, then semantically it doesn't matter so much whether we use a pre-packaged cellular-X as genus (as is the case for GO:0140041 in master), or it's expanded (as in this PR). But it's simplest for us if we use the same structure for everything.

pgaudet:
Hi Chris, 

I dont know that we have a design pattern for this one. All I did was to standardize what we have around that branch: 

The parent "cellular detoxification" has EA: 
detoxification
 and ('occurs in' some cell)

Some children such as 'cellular detoxification of cadmium ion' had 
detoxification
 and ('occurs in' some cell)
 and ('has input' some 'cadmium cation')

while others such as 'cellular detoxification of methylglyoxal' 
had 
'cellular detoxification'
 and ('has input' some methylglyoxal)

Since that gave the same inferences (according to protégé) I made the latter the same as the former. 

The other two terms, 'nitrogen compound' and the new 'fluoride', are not in the ChEBI imports. Should this be part of the same ticket or can the logical definitions be created separately? 

Thanks, 

Pascale

pgaudet:
Hi Chris, 

Let me know if it's OK to merge. 

Thanks, Pascale

cmungall:
OK to merge yes but we need to address the cellular-X DP soon

Issue #13493, opened 2017-05-12
many is_a incomplete in failed Jenkens builds


hdrabkin:
Starting on build 437 (Wed 10th); terms appear to span the ontology, affecting many seemingly unrelated terms

hdrabkin:
I just did a pull and looked at several terms in protege after running the reasoner; there appear to be either asserted or inferred links for those terms. closing this ticket now

hdrabkin:
Still failing; reopening
http://build.berkeleybop.org/job/build-go-ontology/442/

hdrabkin:
fixed apparently

PR #14424, opened 2017-10-24
Adding reduce step back in to pipeline

cmungall:
Should remove redundant links, see #13493


cmungall:
The travis test will pass when https://github.com/ontodev/robot/pull/204 is fixed

Issue #14595, opened 2017-11-20
The term "viral integration complex" cannot be assigned to viral proteins


SIBvirus:
The term "viral integration complex" (GO:0019035) is restricted to eukaryota/cellular organisms. This should not be the case since the viral integration complex is composed of both cellular and viral proteins (cf. GO definition: A nucleoprotein complex containing viral genetic material and associated **viral** and host proteins, which is capable of inserting a viral genome into a host genome).

pgaudet:
Hello, 

Is this term relevant to bacteriophages? I think the problem comes from the fact that the term is under 'nuclear part'; I can move it to 'host cell nuclear part' if this only applies to eukaryotic viruses. 

Thanks, Pascale

pgaudet:
Same for 'viral transcriptional complex'; I presume bacteriophages get transcribed as well.

SIBvirus:
Hello,

The "viral integration complex" is specific to eukaryotic viruses (retroviruses). 

Bacterioviruses get transcribed for sure! I have never seen the term  "viral transcriptional complex" used for bacteriophages though...

pgaudet:
OK. I had put both directly under protein complex. If you want I'll add 'host cell nuclear part'. 

Thanks, Pascale

SIBvirus:
No, it's OK like that, thanks!

pgaudet:
OK perfect, thanks.

PR #14602, opened 2017-11-20
added subclass host cell nuclear part for #14595

pgaudet:


pgaudet:
Changes from the first pull request were OK. I will not merge that second pull request.

Issue #14540, opened 2017-11-09
OTR/NTR: not all synaptic vesicle endocytosis is clathrin-mediated


hattrill:
synaptic vesicle endocytosis GO:0048488 is defined as a clathrin-mediated endocytosis, with all its ancestry.
But it appears that there is at least another flavour of synaptic vesicle endocytosis: activity-dependent bulk endocytosis, which occurs in situations of intense activity.
PMC2871594
https://en.wikipedia.org/wiki/Bulk_endocytosis

It would be good to make the destinction: 
—clathrin-dependent synaptic vesicle endocytosis 

—activity-dependent bulk synaptic vesicle endocytosis
synonym:activity-dependent bulk endocytosis, ADBE, bulk endocytosis
Def: Endocytosis of large regions presynaptic membrane to form endosomes from which new synaptic vesicles are formed. Occurs during periods of intense stimulation. PMID:19266323.

ukemi:
@BarbaraCzub do you want to take this one?

BarbaraCzub:
On my to-do list!

hattrill:
@BarbaraCzub I think that @ukemi has given you a real stinker of an assignment here!

ukemi:
I would:
Read the reference cited above.
Rename the current term to fit its definition. 
Check the part of children of the current term to make sure they are all consistent with clathrin-mediated endocytosis. Initial inspection suggests they are, but maybe could use a few tweaks to definitions. 
Create a new generic parent with the name of the current term
Create the new subtype suggested in this ticket.

Doesn't seem so bad. Am I missing something?

hattrill:
Simple!

BarbaraCzub:
@hattrill I hope there are not hidden surprises in this request! :) @ukemi, please correct me, if I'm wrong, but my understanding is that this request is similar in principle to the types of edits, which we did, when we worked through the dendritic branch during my training in London. So I think I should be ok.

ukemi:
Hi Barbara,

Yes. That is correct.

BarbaraCzub:
PMC2871594  =  PMID:19266323

BarbaraCzub:
@hattrill what GOC xrefs should I include with the new terms?

BarbaraCzub:
### Advice requested re: intracellular transport and (synaptic vesicle) endocytosis 

[...]

PR #14665, opened 2017-11-28
Revised definition, changed parent from clathrin-dependen endocytosis…

BarbaraCzub:
… to endocytosis #14540 

Hi @pgaudet and @vanaukenk 

I have changed the definition of 'synaptic vesicle endocytosis' to:

'A vesicle-mediated transport process, in which presynaptic membrane constituents are retrieved from the synaptic cleft after neurotransmitter secretion by exocytosis. Synaptic vesicle endocytosis can occur via clathrin-dependent and clathrin-independent mechanisms.'

I have also changed the is_a parent from 'clathrin-dependent endocytosis' to just 'endocytosis'. 

There are still other aspects of this ticket that I need to work on (incl. creating the more specific child terms), so I am not closing it yet.

In the meantime, could you please review the new definition and the new parent before I merge?

Thank you,
Barbara


vanaukenk:
@BarbaraCzub - this looks okay to me.

pgaudet:
Looks ok to me too, you can merge. 

Thanks, Pascale

BarbaraCzub:
Hi @pgaudet and @vanaukenk 
Thank you for your comments. Although having slept on this, I now think that the definition I initially suggested is wrong... 

Because (1) the main goal of synaptic vesicle endocytosis is the retrieval of the excess membrane from cell surface after neurotransmitter release, and (2) I do not think anything other than the neurotransmitter gets secreted into the synaptic cleft during the preceding neurotransmitter release (and presumably the neurotransmitter would be fully taken up by the postsynaptic cell), so the synaptic cleft probably should not be mentioned. 

Consequently I would like to suggest this new definition instead:
'A vesicle-mediated transport process, in which the synaptic vesicle membrane constituents are retrieved from the presynaptic membrane on the axon terminal after neurotransmitter secretion by exocytosis. Synaptic vesicle endocytosis can occur via clathrin-dependent and clathrin-independent mechanisms.'

And now my question is: how do I do this? Can I delete this pull request somehow? Or should I merge it now, and then update the definition in a new edit and then pull and merge? I will be adding the child terms today, so I may as well go with the second option (which is probably safer?).

Thanks,
Barbara

pgaudet:
Hi @BarbaraCzub 

You can just close the pull request without merging it, and start over (or correct the branch you were working on). 

Just one point: Is this different from This is really not my area; I can look more deeply but I cannot advise much on th ebiology right now. 

Just, now this really sounds like 'neurotransmitter reuptake' ? 

Thanks, Pascale

BarbaraCzub:
Hi @pgaudet 

>Just, now this really sounds like 'neurotransmitter reuptake' ?
The definition, which I originally suggested did indeed sound too much like neurotransmitter reuptake, so having thought about it more, I revised it to focus on the membrane retrieval. 

>You can just close the pull request without merging it, and start over (or correct the branch you were working on).
Ok, so I assume that if I click 'Close and comment' then this will close my pull request. And then I can go back to editing and change the definition, and this will (presumably) result in a _new_ pull request?

Thanks,
Barbara

BarbaraCzub:
Ok, now I have the option to 'Delete branch'. But because I will be changing the definition, and adding child terms, I suppose I should _not_ delete it?
Or is it better to delete it (so that nobody else will merge it seeing that the travis checks have passed, and thinking that perhaps I forgot)? Or would GitHub not allow me to then open a new branch with the same name as this deleted branch?

pgaudet:
If you want to work on the same branch, don't delete it. 

If you delete it, you should create a new branch with a different name to avoid creating any confusion for GH

PR #14700, opened 2017-12-04
Adding gorel target to Makefile and Rebuilt gorel

cmungall:
Email context:

>> I am giving a talk in a Curation workshop on molecular and causal interactions (on Tuesday). I am planning to talk on the AE field relations. 
>> But I am a bit confused. I thought the relations we can include in the AE field all existed in RO, but this doesn’t seem to be the case, in 
>> that I can’t find ‘regulates transcription of’ the best I can find is ‘increases expression of’
>>
>> If the AE relations are not in RO are they an where other than Protien2GO?
>
> They live in gorel, which  imports RO.  OBO version used by Protein2GO is here
>
> https://github.com/geneontology/go-ontology/blob/master/src/ontology/extensions/gorel.obo 
> <https://github.com/geneontology/go-ontology/blob/master/src/ontology/extensions/gorel.obo>
>
> Master file is here:
>
> https://github.com/geneontology/go-ontology/blob/master/src/ontology/extensions/gorel-edit.owl 
> <https://github.com/geneontology/go-ontology/blob/master/src/ontology/extensions/gorel-edit.owl>
>
> But it seems the Jenkins job that converted from edit to release versions was never pointed at the files on GitHub.  This is a problem for me as 
> I need to update in order to allow fly anatomy extensions for FlyBase.
>

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -352,6 +352,25 @@ mirror/%.owl: mirror/%-download.owl
 	$(OWLTOOLS) $< $(FILTER_EXTERNAL) -o $@
 .PRECIOUS: mirror/%.owl
 
+
+# ----------------------------------------
+# ANNOTATION RELATIONS
+# ----------------------------------------
+
+# Create a module from RO that is seeded from the gorel-edit ontology.
+# Note that to be a seed, a relation needs at least one logical axiom;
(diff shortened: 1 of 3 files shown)


pgaudet:
@balhoff Can you check whether this is resolved ?

Issue #14949, opened 2018-01-29
Large number of relations deleted in go-plus.json - is it correct?


tonysawfordebi:
Over the weekend, our ontology update pipeline has been failing because one of its safety checks has detected the deletion of a suspiciously large number of relations (3131 in total) from go-plus.json.

When I investigated, I found that they are all is_a relations between regulation terms, for example:

- GO:0060542 (regulation of strand invasion) is_a GO:0000018 (regulation of DNA recombination)
- GO:2000805 (negative regulation of termination of RNA polymerase II transcription, poly(A)-coupled) is_a GO:0000122 (negative regulation of transcription by RNA polymerase II)
- GO:0060633 (negative regulation of transcription initiation from RNA polymerase II promoter) is_a GO:0000122 (negative regulation of transcription by RNA polymerase II)
- GO:0050715 (positive regulation of cytokine secretion) is_a GO:0001819 (positive regulation of cytokine production)
- GO:0051893 (regulation of focal adhesion assembly) is_a GO:0001952 (regulation of cell-matrix adhesion)

I can force our pipeline to accept these deletions, but first I want to confirm that they are genuine, and not the result of a problem with the ontology (in which case I won't do anything until the problem is fixed).

ukemi:
Hmm. I wonder if ths is due to #14940. We will investigate and @cmungall, perhaps we need to revert.

ukemi:
Maybe not. It looks like that pull request isn't merged yet.

cmungall:
@ukemi looks like the effect you'd expect of losing the r o part-of property chain, can you experiment with adding this back in to see if the inferences come back, cc @balhoff

balhoff:
I'm reverting all my changes for the time being so that I can do this again more carefully.

tonysawfordebi:
Thanks, @balhoff - I've just tried running our update pipeline again, and it looks like your reversion of the changes hasn't found its way into go-plus.json yet.

ukemi:
Note this also relates to #13926. The property chain will eventually be problematic as we move on to functional causality in GO-CAM models and will need to be removed at that point. I like the suggestion in this ticket of somehow asserting all the existing (or soon to be re-existing) relations and then removing the property chain.

ValWood:
We also lost a lot (~100)  of annotations from our slim last night (or the night before)...could this be the cause?

[...]

PR #14825, opened 2018-01-03
Remove RO logical axioms from GO file and update ro_import.owl

balhoff:
`go-edit.obo` contained axioms that should be provided by RO. I removed all logical axioms from RO terms in `go-edit.obo`, and regenerated `ro_import.owl` using the recently improved `seed.txt`.

The following axioms from `go-edit.obo` are no longer present in the resulting GO release (i.e. they are not in RO):

```
TransitiveObjectProperty(obo:RO_0001000)
SubObjectPropertyOf(obo:RO_0002211 obo:RO_0002328)
SubObjectPropertyOf(obo:RO_0002211 obo:RO_0002410)
SubObjectPropertyOf(obo:RO_0002211 obo:RO_0002418)
SubObjectPropertyOf(obo:RO_0002216 obo:RO_0002595)
SubObjectPropertyOf(obo:RO_0002334 obo:RO_0002328)
SubObjectPropertyOf(obo:RO_0002334 obo:RO_0002410)
SubObjectPropertyOf(obo:RO_0002340 obo:RO_0002313)
SubObjectPropertyOf(obo:RO_0002345 obo:RO_0002313)
SubObjectPropertyOf(obo:RO_0002404 obo:RO_0002410)
SubObjectPropertyOf(obo:RO_0002411 obo:RO_0002410)
SubObjectPropertyOf(obo:RO_0002418 obo:RO_0002410)
SubObjectPropertyOf(obo:RO_0002427 obo:RO_0002410)
SubObjectPropertyOf(obo:RO_0002465 obo:RO_0002464)
SubObjectPropertyOf(obo:RO_0002481 obo:RO_0002464)
SubObjectPropertyOf(obo:RO_0002482 obo:RO_0002464)
SubObjectPropertyOf(obo:RO_0002596 obo:RO_0002595)
```

PR #14958, opened 2018-01-29
Revert removal of RO axioms from go-edit

balhoff:
Temporarily revert PR #14825. Need to more carefully test for loss of inferences. Should address #14949.


balhoff:
@ukemi I'm just reverting my whole pull request now so that I can take more time to figure out where the problems are.

ukemi:
Thanks @balhoff. Let me know if there is anything you would like to look at together.

Issue #14885, opened 2018-01-18
NTR: oxidised low-density lipoprotein particle receptor activity


BarbaraCzub:
PMID:27607416, Figure 5.

Background in #13712

Parent term: GO:0030228 lipoprotein particle receptor activity*
Definition: Combining with an oxidised low-density lipoprotein particle and delivering it into the cell via endocytosis. PMID:27607416, GOC:aruk, GOC:bc
Exact Synonyms: ox-LDL particle receptor activity, oxLDL particle receptor activity, oxidised LDL particle receptor activity, oxidized LDL particle receptor activity, oxidized low-density lipoprotein particle receptor activity
Related Synonyms: oxidised low-density lipoprotein receptor activity, ox-LDL receptor activity, oxLDL receptor activity, oxidised LDL receptor activity, oxidized LDL receptor activity, oxidized low-density lipoprotein receptor activity

*LDL- and oxLDL-related terms should be 'siblings' based on ChEBI
image
![image](https://user-images.githubusercontent.com/16541096/35097058-650979aa-fc46-11e7-88c4-e6887f88e0fa.png)

@paolaroncaglia
@RLovering

RLovering:
Hi Barbara

it would be helpful if you explained what you are proposing here. Are these currently siblings? do these terms exist.

I don't want to have to look up the ontology to try to work out why you are raising a ticket

Thanks

Ruth

BarbaraCzub:
Hi Ruth, 

This is a new term request. I was just reviewing some P2GO tickets pending, and realised that the cellular response term #13712 had been updated from LDL to oxLDL after we discussed this (22nd June 2017 UCL team meeting minutes), but the receptor activity and clearance terms #14884 had not been. They are needed because the paper showed that oxLDL had an effect, but LDL did not. The reference to these terms being siblings is a reminder for me _not_ to make oxLDL a child of LDL, but to follow the same hierarchy as ChEBI, and make them both child terms of the same parent term.

Please let me know, if you have any other questions.

Thanks,
Barbara

paolaroncaglia:
@BarbaraCzub 
This request looks good to me too, thanks.
While checking your suggested parent GO:0030228 lipoprotein particle receptor activity, I noticed that the label of its child GO:0005041 low-density lipoprotein receptor activity should be corrected to low-density lipoprotein _particle_ receptor activity. Could you fix that, please? And keep 'low-density lipoprotein receptor activity' as a related synonym. Many thanks.

[...]

PR #15038, opened 2018-02-06
Added new term and the missing particle as pointed out by Paola Fixes…

BarbaraCzub:
… #14885 

Hi @ukemi the new term should be ok, but I also changed the label of a related term as @paolaroncaglia suggested: "GO:0005041 low-density lipoprotein receptor activity should be corrected to low-density lipoprotein _particle_ receptor activity". Could you please confirm that it is ok to merge this? 
Thanks,
Barbara

cc @RLovering


BarbaraCzub:
oh no. I chose the wrong namespace... should I wait for the pending check to finish before editing this? Or can I edit and push again while this is running?

ukemi:
Shouldn't the namespace be molecular_function?

BarbaraCzub:
I have just corrected this. Yes.

BarbaraCzub:
But I have noticed another error that I made...  I put the pull request number after 'Fixes' instead of the issue number. Is there a way to edit this?

ukemi:
It shouldn't be a problem.

BarbaraCzub:
Ok, thank you @ukemi, I'll merge now.

PR #15059, opened 2018-02-07
NTR complex request per ticket 15021

hdrabkin:
m7G(5')pppN diphosphatase complex


ukemi:
Looks like you fixed this!

hdrabkin:
except that in protege there is a term that does not hyperlink; but I guess travis doesn't care

PR #15060, opened 2018-02-07
adds an equivalent class check, allowing asserted equivalent classes

dougli1sqrd:
This changes the makefile so that when building `reasoned.owl` it checks for non-asserted equivalent classes, and fails (exits) if it finds any. I think this is what you're looking for.


ukemi:
Sigh. Yes these are all real conflicts that need to be corrected. We need to be more careful with our logical definitions. @pgaudet are you going to correct these?

pgaudet:
@ukemi yes I'll correct them right now. @dougli1sqrd hopefully you can test on a different branch ?

dougli1sqrd:
Since you fixed the equivalent classes, my check will pass.

cmungall (approved):

PR #15102, opened 2018-02-12
Update uberon_terms.txt

ukemi:


ukemi:
@cmungall once we have merged this can we regenerate the import file so I can create logical defs for the new terms?

PR #15128, opened 2018-02-13
Update so_terms.txt

ukemi:
Needed to create a logical def for the binding term.


ukemi:
I believe we will also need to run the makefile on this import.

PR #15200, opened 2018-02-19
Added the new term and synonyms Fixes #15188

### LINKED ISSUE #15188: distal axon
I would like to see "distal axon” added to the GO ontology under the parent term "axon part” (GO:0033267).  The definition could be "That part of an axon close to and including the growth cone or the axon terminus.”  Example can be found in PMID:17202468.

BarbaraCzub:


ukemi:
There is no need to add part_of axon if it is already asserted as an axon part. Otherwise the term looks good. Should growth cone and other things be made children of this?

BarbaraCzub:
Thanks @ukemi, I'll remove the 'part_of axon'; and I'll add the part_of relations between the 'growth cone' and the 'distal axon' as well as the 'axon terminus' and the 'distal axon' (as the term definition implies).

BarbaraCzub:
Hi @ukemi. I am not sure what I am doing wrong, but although I added the subclass/relation part_of distal axon to the 'axon terminus' term (screenshot below), the code seems to suggest that I added this relation to the parent of the  'axon terminus' term, i.e. the 'neuron projection terminus'... which is not right, but I do not seem to be able to fix it.
screen shot 2018-02-19 at 15 48 44

PR #15282, opened 2018-02-28
Fixes #15240

### LINKED ISSUE #15240: Add user id to references for ticket 10873
Hi Chris: here's the ticket request

Well now that I have done almost all of the Sue Rhee/Peifen Zang/ metacyc/plant enzyme request (#10873, a couple terms still being fleshed out), I realize that we should have also had Peifen add  GOC:pz from the users.yaml file  as an additional definition reference for all of the terms.  
Is there an easy way you could do this with a script (that is, for a range of GO ids, append GOC:pz to the definition references? That way it’s easy to identify the terms.
 There are currently about 51 occurrences of GOC:pz in the current file already; these appear to be terms that Tanya made but used GOC:pz as the reference.
 The id range is something like
GO:0102016  through GO:0103118 (~1102; not all ids were used in the end due to rejections and those found to be duplicates of already existing terms):

cmungall:
added goc:pz

Fixed using obo-sed

I did it like this:

```
obo-sed.pl -R GO:0102016 GO:0103118 's@(def: .*)\]\n@$1, GOC:pz\]\n@' go-edit.obo 
```


hdrabkin (approved):
I looked at 50; this is exactly what I wanted to happen; looks good

hdrabkin:
This is all set to merge?

ukemi:
As long as we don't require a new line at the end of the file, this looks good to me. That's why the deleted number of lines doesn't exactly match the added number.

hdrabkin:
No, shouldn't need a new line at end of file I would think

hdrabkin:
Not sure: the current obo file has a couple of blank lines at the end of the file.
right after
[Typedef]
id: transports_or_maintains_localization_of
name: transports or maintains localization of
namespace: external
xref: RO:0002313
is_a: has_participant ! has participant
is_a: related_via_localization_to ! related via localization to
transitive_over: has_part ! has_part
>>two blank lines

pgaudet:
@hdrabkin Can you resolved the conflicts and merge ?

hdrabkin:
Hi @cmungall ; I'm thinking now that the file is so old relative to current that trying to resolve will be horrendous; would it be possible to just rerun the script  and repush?

PR #15284, opened 2018-02-28
outputing times for each command automatically

dougli1sqrd:
In the service of understanding which commands take the longest _on travis_ 

Locally (on my computer), the travis_check does take just 10 minutes. If we see something that takes a super long time, we'll see it in the travis logs.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -3,6 +3,7 @@ all: stage_release
 # ----------------------------------------
 # CORE VARIABLES
 # ----------------------------------------
+SHELL = ./report_time.sh
 
 # base URI for all OBO library ontologies
 OBO=http://purl.obolibrary.org/obo
--- a/src/ontology/report_time.sh
+++ b/src/ontology/report_time.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+shift
+time sh -c "$*"


cmungall:
Which step? I assume reduce. We should have robot produce metadata in
header including time to produce. There is a ticket on the robot tracker
about this, using prov

On Feb 28, 2018 15:30, "Eric Douglass" <notifications@github.com> wrote:

ROBOT_JAVA_ARGS=-Xmx12G robot reason -i enhanced.owl -r ELK -e
asserted-only relax reduce annotate -V http://purl.obolibrary.org/
obo/go/releases/date +%Y-%m-%d/go.owl -o reasoned.owl Is the longest
command. In the makefile, this is make reasoned.owl

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/geneontology/go-ontology/pull/15284#issuecomment-369389702>,
or mute the thread

dougli1sqrd:
In the Makefile, this is `make reasoned.owl`. 

But when a PR is especially long, I'd like to see how long that took too. Although 20 minutes is a little sad for a travis check, I'm more curious about >30 min

dougli1sqrd:
screen shot 2018-03-12 at 1 49 37 pm
It looks like `reduce` is the largest culprit, taking up 6 of the 7 minutes for the reasoned command: `ROBOT_JAVA_ARGS=-Xmx12G robot  reason -i enhanced.owl -r ELK -e asserted-only relax reduce  annotate -V http://purl.obolibrary.org/obo/go/releases/`date +%Y-%m-%d`/go.owl -o reasoned.owl`

cmungall:
I will look into making reduce more efficient.

For now I will alter the pipeline so that reduce is bypassed during travis

dougli1sqrd:
Okay. Let me know if you want me to do anything more, on robot or in the makefile.

PR #15350, opened 2018-03-08
Added syns and xrefs for ketolysis

cmungall:


cmungall:
Need a biologist review - @ukemi @hdrabkin @deustp01 ?

ukemi:
Ha. I was just looking at Reactome for the other ketone body ticket. Seems all reasonable to me and the Reactome xref looks correct. Aligning the Reactome and Metacyc pathways looks reasonable, so that xref is appropriate. And just as an aside, the GO-CAM model for this looks pretty nice!

PR #15370, opened 2018-03-09
Issue 15249

pgaudet:


cmungall:
Here is the problem:

https://github.com/geneontology/go-ontology/pull/15370/files#diff-3cf7b2b4f9db4df4cae7c72388fdba78R112728

cmungall:
Unfortunately you have gone beyond the event horizon here, you've saved a file that is impossible to read again, the only fix is to hand edit file (or redo the work).

This is currently an undesirable feature of the owlapi obo parser/writer (which I wrote). What is written should always be readable, but sadly there are edge cases like this. It's hard to get around this as it arises from a property of .obo.

As a workaround, we'd like better lint checks within Protege, as it's easy to accidentally slip and use the wrong AP for something. I think it should be straightforward to write a robot plugin for Protege: https://github.com/ontodev/robot/issues/265

pgaudet:
How dd you find that ? I think I made that mistake in the past, I don't remember it being so lethal.

cmungall:
When I ran locally I scanned the ginormous OWLAPI error log. the trick is to look for the OBOFormat section (I have done this many times for many different projects... paying the sins for .obo format). It gives the line number. I think it should have been there in the travis log too

PR #15483, opened 2018-03-27
Announcing changes to curators

pgaudet:
Emails for announcing changes to the ontology

### DIFF
--- a/docs/Announcing changes to curators
+++ b/docs/Announcing changes to curators
@@ -1,0 +1,35 @@
+Term obsoletion 
+
+Standard mail: 
+Subject line: Obsoletion notice: GO:nnnnnn [class label]
+Body: 
+Dear all, 
+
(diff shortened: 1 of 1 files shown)


pgaudet:
I integrated these suggestions in this page: https://github.com/geneontology/go-ontology/blob/master/docs/ObsoleteTerm.md

ukemi:
Hi Pascale,

This looks good, but keep in mind the plan is to move away from the markdown file and eventually have the official documentation on the wiki. In my most recent obsoletion e-mail, I added Gene Ontology to the subject line. I also cc'd the contacts for the other ontology that were listed as using the term according to ontobee. Did we decide that we weren't going to do that anymore?

ValWood:
>  I also cc'd the contacts for the other ontology that were listed as using the term according to ontobee.

this would be useful for us

PR #15504, opened 2018-03-29
adding synonym label matching violations

dougli1sqrd:
Finds pairs of classes where one class's label is the same as another's synonym. 

Question: Once these are detected, how does one repair this? Which class is "at fault" for duplicated labels/synonyms?


ukemi:
Shouldn't this only be done when the synonym scope is exact? I think it is valid to have the same string for synonyms of other scope.

ukemi:
https://github.com/geneontology/go-ontology/issues/15686

dougli1sqrd:
Thanks

Issue #15429, opened 2018-03-16
replaced_by:obsolete term id situations


tberardini:
Our OBO parsing scripts have been failing on loading because some obsolete terms have replaced_by terms that are themselves obsolete or were merged into other terms.  Upon closer examination, there are only a few but they are doozies.

(1) GO:0022815 replaced by GO:0022891 which was itself replaced by GO:0022857
(2) GO:0003733 replaced by GO:0030529 which was merged into GO:1990904
(3) GO:0030531 replaced by GO:0030529 which was merged into GO:1990904
(4) GO:0030356 replaced by GO:0030531 which was replaced by GO:0030529 which was merged into GO:1990904

I suggest that the replaced by fields be updated as follows:

(1) GO:0022815 replaced by GO:0022857
(2) GO:0003733 replaced by GO:1990904
(3) GO:0030531 replaced by GO:1990904
(4) GO:0030356 replaced by GO:1990904

Any objections?  How can we prevent this from happening in the future?  Is this a problem that others have also encountered in their GO loading pipelines?

cmungall:
@dougli1sqrd will implement a check to prevent this happening moving forward.

I'll repair these in place for now

tberardini:
Thank you, Chris!

dougli1sqrd:
I made a sparql query that finds a couple more. Let's make sure that this isn't being to aggresive:
```
?class	?replacement
<http://purl.obolibrary.org/obo/GO_0008581>	<http://purl.obolibrary.org/obo/GO_0004221>
<http://purl.obolibrary.org/obo/GO_0008436>	<http://purl.obolibrary.org/obo/GO_0030530>
<http://purl.obolibrary.org/obo/GO_0030356>	<http://purl.obolibrary.org/obo/GO_0030531>
<http://purl.obolibrary.org/obo/GO_0003733>	<http://purl.obolibrary.org/obo/GO_0030529>
<http://purl.obolibrary.org/obo/GO_0022815>	<http://purl.obolibrary.org/obo/GO_0022891>
<http://purl.obolibrary.org/obo/GO_0030531>	<http://purl.obolibrary.org/obo/GO_0030529>
```

The sparql for this was: 
```
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX IAO: <http://purl.obolibrary.org/obo/IAO_>

SELECT ?class ?replacement
WHERE {
  ?class IAO:0100001 ?repl .
[...]

PR #15506, opened 2018-03-30
check that detects replaced by classes that are deprecated but not merged

dougli1sqrd:


tberardini:
Failing checks seems bad omen.

ukemi:
I think this is ok and the checks are working. See my comments on the ticket #15429

dougli1sqrd:
@ukemi @tberardini does this work?

ukemi:
Yes. I had actually already made this fix by hand in a branch and not pushed, but what I was asking in the ticket is whether or not we wanted to always fix these by hand or whether these errors could be handled programmatically. Sorry if I wasn't clear.

PR #15583, opened 2018-04-17
Added term to goslim_synapsesubset Fixes geneontology/synapse#179

BarbaraCzub:
Hi @ukemi 
I've just attempted to add one of the requested terms to a SynGO slim. Could you please confirm I've done this correctly?

My attempts to follow these instructions were not succesful: 
http://go-ontology.readthedocs.io/en/latest/NewSlimTerm.html
(I tried from different starting points several times, but 'git diff' never showed anything).
cc @vanaukenk 

So instead I just clicked the '+' button next to 'Subset' in the 'OBO annotation' window in Protege, and _typed in_ goslim_synapse, which seems to have done the job.
(Please see line 462508 in the file changed here and compare to e.g. line 441200 in https://github.com/geneontology/go-ontology/pull/14651/files).

Is this ok?

Thanks,
Barbara


ukemi:
Hi @BarbaraCzub 

Can you skype?

PR #15631, opened 2018-04-20
Added requested term, corrected logical definition of a related term …

BarbaraCzub:
…Fixes geneontology/synapse#195

cc
@RLovering 
@thomaspd


ukemi (approved):
I'm not sure why line 459724 disappeared and then came back, but this all looks ok to me.

PR #15639, opened 2018-04-23
Added requested term Fixes geneontology/synapse#184

BarbaraCzub:
cc
@thomaspd 
@RLovering


ukemi (changes requested):
Term 150047 is a function, but it is listed as a process.

+[Term]
+id: GO:0150047
+name: G-protein coupled neurotransmitter receptor activity involved in regulation of presynaptic membrane potential
+namespace: biological_process

ukemi (approved):

PR #15643, opened 2018-04-23
Added requested term Fixes geneontology/synapse#206

BarbaraCzub:
cc
@thomaspd 
@RLovering


ukemi (commented):
I'm not sure the has_part 'x cell' relationships are appropriate. Does the synapse have_part the whole cell? Is this a pattern that you got from other terms?

BarbaraCzub:
Hm... No, it is not the whole cell. So, what should the relationship be? 'Overlaps'?

BarbaraCzub:
And can I use 'overlaps' in a logical definition, or should I add these relationships in the 'SubClass Of' field instead?

ukemi:
Are there any other examples that are already in place?

PR #15656, opened 2018-04-26
Added requested term Fixes geneontology/synapse#209

BarbaraCzub:
cc
@thomaspd 
@RLovering


ukemi (commented):
Based on the definition, should it be an is_a septin cytoskeleton or a part_of septin cytoskeleton?

BarbaraCzub:
Well, this is exactly what I was not sure about, hence my request for review...
I decided to go for is_a because this is what was done in similar cases. E.g.
screen shot 2018-04-26 at 17 16 00

BarbaraCzub:
Ok... It appear's that I opened another Pandora's box by requesting this review (sorry!).

I based the definition of 'postsynaptic septin cytoskeleton' on definitions of similar terms that are related by is_a, e.g. like in the 'postsynaptic actin cytoskeleton' above. Or 'postsynaptic cytoskeleton' is defined as: 'The portion of the cytoskeleton contained within the postsynapse' (is_a 'cytoskeleton'). 

So revising the textual definition of this term would require a revision of the other definitions too. 

Perhaps I could merge now as it is, and we can add this as a discussion point to one of the ontology calls? 

Because I do not even know how best to re-phrase this to not make is sound like part_of... 
"The portion of the septin cytoskeleton that overlaps with the postsynapse."?
Or the other way round:
"The postsynaptic cytoskeleton that consists of septins and associated proteins."?

ukemi:
😃 Sound ok to me.

BarbaraCzub:
You mean one of the alternative definitions sounds good, or merging now as it is sounds good?

ukemi:
Keep it consistent with the other synapse terms.

BarbaraCzub:
Thank you!

PR #15717, opened 2018-05-10
Added term to goslim_synapse Fixes geneontology/synapse#217

BarbaraCzub:


BarbaraCzub:
cc
@RLovering 
@thomaspd 
@ukemi

PR #15738, opened 2018-05-13
Dup synonyms issue 13667

dougli1sqrd:
I should also add Kimberly,  but I don't know her github name...


ukemi:
Hi @dougli1sqrd,
I don't think we want to restrict this to just exact synonyms. The helpdesk ticket that triggered this pointed out two synonyms that were exactly the same string, but one was narrow and the other exact.

ukemi (changes requested):
Hi @dougli1sqrd,
I don't think we want to restrict this to just exact synonyms. The helpdesk ticket that triggered this pointed out two synonyms that were exactly the same string, but one was narrow and the other exact.

cmungall:
These should be two distinct tests:

 * no two classes should share a single exact synonym
 * no one class should share the same string as both a X and Y scope synonym (e.g. X=related, Y=exact)

Note: @rctauber is integrating the first test and I think optionally the 2nd into robot core, since we're all here we should coordinate

ukemi:
Yes. Sorry. I thought the exact synonym issue was a separate ticket.

ukemi:
Errors should be fixed in #15744

ukemi (approved):

Issue #15632, opened 2018-04-20
NTR: rRNA pseudouridine synthase activity


keseler:
For some reason, the only children of GO:0009982 are tRNA and snRNA pseudouridine synthase activities.  Could you please add the corresponding rRNA term?  Proposed analogous definition:
Catalysis of the reaction: rRNA uridine = rRNA pseudouridine. Conversion of uridine in a tRNA molecule to pseudouridine by rotation of the C1'-N-1 glycosidic bond of uridine in RNA to a C1'-C5.

EC numbers for these enzymes are 5.4.99.16 through 5.4.99.24.

Also, for some reason the "general" GO:0009982 has all those tRNA-related synonyms; maybe those should be moved to the tRNA-specific term, GO:0106029?

krchristie:
Regarding relevant EC #'s for pseudouridine synthase activities, using both of these websites, I see these pseudouridine synthase activities:
- http://www.sbcs.qmul.ac.uk/iubmb/enzyme/EC5/cont5a.html
- https://enzyme.expasy.org/cgi-bin/enzyme/enzyme-search-ec

**5.4.99.12 - tRNA pseudouridine(38-40) synthase**
5.4.99.25 - tRNA pseudouridine(55) synthase
5.4.99.26 - tRNA pseudouridine(65) synthase
5.4.99.27 - tRNA pseudouridine(13) synthase
5.4.99.28 - tRNA pseudouridine(32) synthase
5.4.99.42 - tRNA pseudouridine(31) synthase
5.4.99.45 - tRNA pseudouridine(38/39) synthase

5.4.99.44 - Mitochondrial tRNA pseudouridine(27/28) synthase

5.4.99.19 - 16S rRNA pseudouridine(516) synthase

5.4.99.20 - 23S rRNA pseudouridine(2457) synthase
5.4.99.21 - 23S rRNA pseudouridine(2604) synthase
5.4.99.22 - 23S rRNA pseudouridine(2605) synthase
5.4.99.23 - 23S rRNA pseudouridine(1911/1915/1917) synthase
5.4.99.24 - 23S rRNA pseudouridine(955/2504/2580) synthase
5.4.99.29 - 23S rRNA pseudouridine(746) synthase

5.4.99.43 - 21S rRNA pseudouridine(2819) synthase

[...]

PR #15747, opened 2018-05-14
more synonyms & xrefs from EC for tRNA & rRNA pseudouridine synthases…

krchristie:
…, #15632


hdrabkin:
Yes, that is what I'm thinking we should do to clarify. In the meantime, maybe you should merge this so it doesn't get too far out of sync, and then if it needs to be changed it can be done after the Editors call?

krchristie:
Sounds like a good idea to merge sooner rather than later, so I'll go ahead and change the synonym that I thought maybe should also be EXACT to be that instead of NARROW, and then merge. If we decide as a group we shouldn't have multiple narrow xrefs, I'll take them out.

PR #15763, opened 2018-05-21
DO NOT MERGE YET: Rebuild PR import.

balhoff:
@cmungall I built an update for `pr_import.owl`. It would be good for you to review, as there are substantial changes, and lots of new `HGNC` terms (which doesn't seem to be a valid OBO namespace). Thoughts?


balhoff:
I am redoing this.

PR #15766, opened 2018-05-21
Rebuild RO import.

balhoff:
@cmungall @vanaukenk any reason to hold off on updating RO? Fairly high number of changes here.


cmungall (approved):
+1

also I'm making a more componentized ro release to avoid the pato/caro in ro import
https://github.com/oborel/obo-relations/pull/231

PR #15896, opened 2018-06-11
Update chebi_import after updating substance_by_role.

balhoff:
The recent failures of build-go-ontology on Jenkins seem to be resulting from failing to write go_inferences.obo. That failure appears to have been the result of an OWL RDF problem in chebi_import.owl with a bad blank node reference. Rebuilding chebi_import did not fix it, but did if substance_by_role was rebuilt first.

### DIFF
--- a/src/ontology/imports/chebi_import.obo
+++ b/src/ontology/imports/chebi_import.obo
@@ -5737,6 +5737,12 @@ relationship: has_role CHEBI:75771 ! mouse metabolite
 relationship: has_role CHEBI:76971 ! Escherichia coli metabolite
 
 [Term]
+id: CHEBI:17623
+name: cyclomaltodextrin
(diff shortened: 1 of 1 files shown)


ukemi:
Could this have been caused by the edit we made last week in PR #15855? I have been wracking my brain trying to figure out the dependencies and why the builds are failing. If we rebuild ChEBI import, will it reintroduce the obsolete GO reference in the import? I'm still having a hard time grasping the whole import pipeline and what happens when. Thanks for looking onto this @balhoff .

balhoff:
Yes that was when the problem occurred, but I don't really understand where `<rdfs:subClassOf rdf:nodeID="_:genid20457"/>` came from. That was the problematic part. But after building a fresh substance_by_role, the GO-to-GO relationship seems to have disappeared.

PR #15902, opened 2018-06-12
169 changes in GOCHE term names due to loading and saving in Protege

krchristie:
@cmungall @ukemi - Do you know why there were so many changes in GOCHE term names? I was trying to do a simple definition change on one term, and decided to commit all these GOCHE term name changes separately from my other work so that the diff isn't swamped by this stuff that changed only upon saving out from Protege.


ukemi:
Hmm. They had changed a few weeks ago too. I wonder if when @balhoff regenerated the file they changed back?

balhoff:
I did just regenerate chebi_import. Are these GOCHE labels just automatically derived or were they supposed to be overridden to those values?

ukemi:
Good question. I thought they were automatically derived, but I am not sure.

Issue #13533, opened 2017-05-24
NTRs: child terms of GO:1903440 amylin receptor complex


BarbaraCzub:
Hello, 

I am currently annotating PMID:22500019, which investigates signalling pathways that 'amylin receptor complex 3 (AMY3)' is involved in. 'Amylin receptor complex 1 (AMY1)' and 'amylin receptor complex 2 (AMY2)' are studied as controls.

I would like to request new GO terms for these three cellular components:
GO NTR1: Amylin receptor complex 1 (AMY1);
GO NTR2: Amylin receptor complex 2 (AMY2);
GO NTR3: Amylin receptor complex 3 (AMY3).
All three will be child terms of GO:1903440 amylin receptor complex.

Please use the DbxRef: ARUK-UCL and GOC:bc.

The Complex Portal entries provide further details:
AMY1: https://www.ebi.ac.uk/intact/complex/details/EBI-9008682
AMY2: https://www.ebi.ac.uk/intact/complex/details/EBI-9685417
AMY3: https://www.ebi.ac.uk/intact/complex/details/EBI-9685439

Thank you,
Barbara

BarbaraCzub:
Similarly, since the paper shows that the different amylin receptor complexes are regulated by different ligands (e.g. supplementary figure 2), could I also please request the following
child terms of GO:0097647 amylin receptor signaling pathway:

GO NTR4: amylin receptor 1 (AMY1) signaling pathway;
GO NTR5: amylin receptor 2 (AMY2) signaling pathway;
GO NTR6: amylin receptor 3 (AMY3) signaling pathway.

Please use the DbxRef: ARUK-UCL and GOC:bc.

Thanks.

bmeldal:
Barbara, we have all of these in the CP already. Unless you can identify specific functions that differ between the 3 forms of the amylin complex we shouldn't create new GO terms for these compositional variants. At least that was the last decision we made.

BarbaraCzub:
Thanks @bmeldal, PMID:22500019 shows that the 3 complexes (AMY1, AMY2, and AMY3) are bound by different ligands (main figures re AMY3 and Supplementary Figure 2 re AMY1 and AMY2). The downstream signalling events are investigated in detail only for AMY3, as this is the receptor, which is shown in this paper to be regulated by amyloid-beta oligomers (AMY1 and AMY2 are not affected by amyloid-beta oligomers according to this paper). 

[...]

PR #15919, opened 2018-06-19
Incorrect parent term removed

BarbaraCzub:
cc
@vanaukenk 
@RLovering


BarbaraCzub:
Re:
https://github.com/geneontology/go-ontology/issues/13533

PR #15921, opened 2018-06-20
Updated Reactome xref for pyruvate kinase GO:0004743

ukemi:

PR #15985, opened 2018-07-02
updated Reactome xrefs

goodb:
Using mapping file provided by Reactome, executed a script that replaced the old REACT_ ids with their corresponding current HSA_..  ids.  Also removed xrefs to non-human reactome ids as these are not stable and are generated automatically from the human ids.  Removed old xrefs to REACT_ids that have no current mapping.  See issue comments https://github.com/geneontology/go-ontology/pull/15921 for more information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ target/
 # Mac specific weirdness
 .DS_Store
 src/.DS_Store
+src/ontology/imports/catalog-v001.xml


ukemi:
Note to @deustp01 to remind us that we need to have a look at the Reactome labels in this branch. Yesterday doing QC we thought that they might not be current.

ukemi:
Today we spot checked the labels and they all looked ok, but should double-check with @goodb to be sure that when he updated the xrefs, he updated the labels also.

goodb:
@ukemi the labels that are attached to database_cross_reference property for the values I updated came from the original GO file, not from the mapping file provided by Reactome.  e.g., for GO:0042284, Reactome:R-HSA-428259, the label is 
dihydroceramide + NAD(P)H + H+ + O2 => ceramide + NAD(P)+ + H2O, Homo sapiens

It has that ", Homo sapiens" on the end because it came from the previous file.  I'll go ahead and regenerate a version with the labels from the Reactome mapping file.

goodb:
@ukemi I pushed another version with the labels updated here.  It looks like Travis doesn't like it and I don't understand why...  Let me know if there is something else I need to do...

ukemi:
It looks like there is a non-ascii character in "CYP27A1 27-hydroxylates 5β-CHOL3α,7α,24(s)-triol". Probably the beta and alpha but could also be the dashes.

ukemi:
How hard would it be just to generate a new fresh version? It looks like there are merge conflicts as well. I suspect with the time this has spent out there, the conflicts are mounting.

ukemi:
@deustp01, should we keep the appended , Homo sapiens?

goodb:
Hi both.  Its not hard to regenerate.  What do you want me to do about the non-ascii problem?  I'm pretty sure I could find a way to filter them out or map to ascii if that is important.  

The most recent commit was from a fresh copy of the ontology from master.  Not sure if you are trying to merge that with master or with the previous commit in this branch?  

Suggest that (1) you let me know what the ascii requirements are here and then (2) I make a new branch off of master, and add in the new ontology file.

deustp01:
Filtering non-ASCII - my non-developer concern is the effect, if any on searches. Simply replacing each non-ASCII character with a question mark may look odd but is acceptable, I think - the issue is whether that replacement or a more aesthetically pleasing one would keep someone looking for things by name from finding them.

ukemi:
I think we have standard way of representing the non-ascii characters. Is there any way to tell how many terms fail the ascii check? I can ask @dougli1sqrd if he can figure this out. If it is only a few, I propose that we close the current pull request and delete the branch. Ben can regenerate the new file and commit it to a new branch. We will create a new pull request that will fail the ascii check. We can fix those in Protege and recommit. If there are a lot of terms, maybe @cmungall has some code stashed somewhere that will replace them programmatically according to GO standards. 

I will discuss this strategy on the editors' call on Monday. @deustp01 is on vacation for a couple weeks, so there is no rush in committing this. We will want to sanity check it when he gets back. Once Peter is back, if we think there has been enough lead time I suggest we go for it.

goodb:
Any resolution on the ascii question?  I can tell you how many of them there are if that helps.  Seems like something there would be a policy on though.

cmungall:
Historically we have always disallowed ascii, and have spelled these out

open to changing this policy. We still want to make sure we do things 
like filter things like non-true-hyphen-hyphen-characters as they can 
confuse exact matches etc

goodb:
It looks like this is the only line in the entire file that contains any non-ascii..  and its just the β and the α chars.

ukemi:
@goodb . I think we can close this?

ukemi:
When I look at the current ontology file, the ...HSA... xrefs are there and the original issue is closed. I think this pull request is outdated, but didn't want to close it without @goodb 's  approval.

goodb:
Yes we can close this one.  This has been superseded by more recent code changes (that pull the xrefs dynamically during the go build).

PR #16011, opened 2018-07-06
Fixing capitalization on synonyms and labels.

cmungall:
Note that even if an abbreviation exists for a term, it does not
follow that the spelled-out version should be capitalized.

Examples:
 - https://www.google.com/search?q=%22Epidermal+Growth+Factor%22
 - https://www.google.com/search?q=%22%22Broad-Complex%2C+Tramtrack+and+Bric+a+brac+domain%22
 - https://www.google.com/search?q=%22Endosomal+Sorting+Complex+Required+for+Transport%22


ukemi (approved):
I agree. Capitalization should only be used for proper nouns.

PR #16174, opened 2018-08-01
uses robot to dump an inconsistent ontology if in go-gaf.owl

dougli1sqrd:


cmungall:
This will be triggered for every PR. Is this your intention?

kltm:
@cmungall Speaking for @dougli1sqrd , I do believe intention was for this to be triggered on every PR (I cannot speak to how resource-hungry that op is). It would not ever cause a fail, just an additional file.

cmungall:
My mistake: travis runs `travis_test` for this repo, which is a subset of `test`, so it does not get executed on PRs

(it would be pointless to do an always-succeed test on travis unless we were to set up a deploy hook)

PR #16239, opened 2018-08-13
Change go-lego ECO import to eco-base.

balhoff:
This will prevent import of OBI axioms which cause trouble until handling of SO terms as generic dependents is improved.


balhoff:
Thanks @rctauber!

PR #16242, opened 2018-08-13
created new term, changed subclasses as requested fixes #15939

### LINKED ISSUE #15939: MP  plasma membrane fusion
GO:0045026 plasma membrane fusion
https://www.ebi.ac.uk/QuickGO/term/GO:0045026

missing parent

GO:0007009 plasma membrane organization
https://www.ebi.ac.uk/QuickGO/term/GO:0007009

Isn't the reasoner working?

pgaudet:


ukemi:
Also, how does fertilization fit in here?

ukemi (changes requested):
See comments.

mah11:
> Is the karyogamy term really a subclass of cell-cell fusion

NO. Karyogamy is fusion of _nuclei_, not cells.

pgaudet:
Following the feedback on the Pull request, I noticed the following missing children for cell-cell fusion: 

'parasexual conjugation with cellular fusion'
'conjugation with cellular fusion'
---
 
WRT 'karyogamy involved in conjugation with cellular fusion': 
- I removed the is_a relation (now we only have 'karyogamy
 and ('part of' some 'conjugation with cellular fusion')'
- Perhaps the term would be clearer if the label was 'karyogamy during in conjugation with cellular fusion' ? @ValWood what do you think ? 

Thanks, Pascale

ValWood:
> Perhaps the term would be clearer if the label was 'karyogamy during in conjugation with cellular fusion' 

Well, it is more human readable, but we do have "involved in" in  other terms when these are process sub parts which need to be disambiguated from a general term.

I'm not sure if we should say "during", because  this really only means "happens at the same time", when really it IS involved_in /part_of. I think in this case the loss in readability is better  than the loss of precision.

PR #16287, opened 2018-08-22
WIP Issue 16283 hyphens

pgaudet:


cmungall:
obo syntax error presumably through hand editing:

https://github.com/geneontology/go-ontology/pull/16287/files#diff-3cf7b2b4f9db4df4cae7c72388fdba78R88393

PR #16419, opened 2018-09-25
Po import 20180925

ukemi:


cmungall:
looks like this is revealing some longstanding issues, etc

```
id: GO:0046844
name: micropyle formation
namespace: biological_process
def: "Establishment of the micropyle, a single cone-shaped specialization of the chorion that allows sperm entry into the egg prior to fertilization." [ISBN:0879694238]
is_a: GO:0003006 ! developmental process involved in reproduction
is_a: GO:0032502 ! developmental process
is_a: GO:0048646 ! anatomical structure formation involved in morphogenesis
intersection_of: GO:0048646 ! anatomical structure formation involved in morphogenesis
intersection_of: results_in_formation_of PO:0020025 ! plant ovule micropyle
relationship: part_of GO:0007306 ! eggshell chorion assembly
```

ukemi:
Interesting unsatisfiable classes after rebuilding this import. We should look at this together on an editors' call.

ukemi:
Yes. But his is actually good because it is identifying longstanding issues that are uncovered by importing external ontologies.

ukemi:
I think it would be a good exercise to go through the error report, loading into Protege, identifying unsatisfiable classes and looking at the explanation panel. This is a nice teaching example.

ukemi:
Looks like there is some confusion with respect to spore-bearing structures and some conflation of anatomy terms and logical definitions with respect to sporangium development and spore-bearing structure development. This is waaaaay out of my field of expertise, but seems like we need a parent sporangium to encompass many taxa that can bear spores (there is one in the PO and one in the FAO and according to GO there is also an oomycete one). This leads me to believe that the GO logical defs are way too specific. I'm tempted just to remove the logical defs for spore-bearing structure development and sporangium development because there is no good anatomy term that makes the def necessary and sufficient for all taxa. The micropyle one is easy. We just pointed to the plant structure instead of the GO CC. @tberardini any advice on the sporangium terms?

ukemi:
And do you know the difference between a spore-bearing structure and a sporangium? The defs are very close to being the same.

tberardini:
> I'm tempted just to remove the logical defs for spore-bearing structure development and sporangium development because there is no good anatomy term that makes the def necessary and sufficient for all taxa. 

Yes to this. 

> And do you know the difference between a spore-bearing structure and a sporangium? The defs are very close to being the same.

I'd need to do some background reading to be sure.  Removing the logical defs seems the cleanest and fastest way to resolve the import clashes at this point in time.

ukemi:
Works for me. We can merge later if we think that is best.

PR #16508, opened 2018-10-13
Implemented DOS-DP patterns for taxon constraints.

balhoff:
This is how taxon constraints edited in TSVs could work.


balhoff:
One issue is that a few of the taxon constraints in the ontology have axiom annotations with sources, e.g.:

`relationship: only_in_taxon NCBITaxon:33208 {source="PMID:12382326"} ! Metazoa`

There aren't that many of these. But I don't think we can have optional axiom annotations in DOS-DP at the moment.

cmungall (approved):

balhoff:
@cmungall in order to produce the `never_in_taxon` annotation, we will need to add IRI-valued annotation properties to DOS-DP.

cmungall:
Do you think we should continue to do it this way? The annotation 
assertion is expanded to a disjointness axiom for reasoning. We could 
just expand to this directly. Or use a in-taxon NOT X structure as the 
direct representation.

balhoff:
I am generating both of those directly: https://github.com/geneontology/go-ontology/blob/2e06ad45310210ec165b8c6227169c8592df6ff5/src/taxon_constraints/never_in_taxon.yaml

I thought there was a concern that some consumers may be using the annotation directly. If that's not true I would be glad to go ahead and merge.

balhoff:
I implemented this in dosdp-tools: https://github.com/INCATools/dosdp-tools/pull/65

Need to propose a change to the schema before merging that implementation.

balhoff:
This is just about ready to go, but now I want to clean up the taxon axiom handling in the makefile. Taxon axioms are coming from a few different places:

- `imports/go-taxon-groupings.owl` — this file includes some defined taxon unions, but it also includes a bunch of other subclass axioms that should be coming only from NCBITaxon; this file also imports `taxslim.owl`.
  - I propose removal of all axioms except for taxon union definitions, and removal of the import.
- `$OBO/ncbitaxon/subsets/taxslim.owl` — this file is directly imported and not produced by the GO makefile. It includes more than 13000 classes.
  - I propose this file no longer be imported. Needed NCBITaxon classes should be obtained through the `imports/ncbitaxon_import.owl` module extraction from referenced seed terms.
- `imports/ncbitaxon_import.owl` — this file is created by module extraction from the full taxonomy from referenced seed terms. Then `owltools --create-taxon-disjoint-over-in-taxon` is applied to create disjointness GCIs.
- `imports/go_taxon_constraints.owl` — this is a new file created via DOS-DP.
  - Need to ensure taxa referenced in this file are added to the seed for `imports/ncbitaxon_import.owl`.

@cmungall am I missing anything? Do we need `taxslim.owl`?

balhoff:
With this update, GO-plus will contain 154 taxa rather than 13679...

PR #16546, opened 2018-10-31
TO BE DISCUSSED: Removing property chain

cmungall:
Removing property chain `transports_or_maintains_localization_of o part_of -> occurs_in`

This was not in RO, it's injected via GO (we should check we are not injecting more)

It was causing unusual inferences. We propagate over occurs_in, so annotations to SVT were being grouped under cytoplasm. (found by @lpalbou)

See explanation:

image


ukemi:
I think we are injecting other property chains. Have a look at the regulates relations. I'm adding this for discussion on an editors' call.

cmungall:
OK good. We should remove this or move them to RO.

This is particular one is an interesting case. At first I thought it was
biologically bogus but not so sure now

balhoff:
You also fixed this in #17195, which I just merged. So I'll close this. I think everyone was in agreement to get rid of it.

PR #16597, opened 2018-11-09
Reactome xrefs import

goodb:
This request handles bringing in Reactome xrefs from an externally provided mapping file.  It uses DOSDP to generate an ontology that contains the xrefs and then imports the ontology into the GO.  Could use touch ups relating to handling of non-ascii chars in import file (it just deletes them thought there are only a few) and most likely there is better Makefile logic that could be written.  But it seems to work and we need it done soon so.. here it is.


ukemi (dismissed):
I couldn't load the ontology from this branch into Protege.

goodb:
@goodb did you run the Makefile with Docker?  The import would need to be generated that way until this is executed on the production server and thus could be downloaded directly from there when you load Protege.

ukemi:
No. That would explain why the file isn't there.

balhoff:
@goodb the way most of the imports work is that they are kept in version control, and periodically regenerated manually. So if you did it that way, the local file would be loaded via the catalog XML.

goodb:
@balhoff the makefile will regenerate the import whenever it is run as it stands now.  Yes, I edited the catalogue.xml file to point to its local version and yes, that will report a problem before the make is made.  It is different from the other imports. Perhaps there is a different pattern that would make more sense to the editors.

ukemi:
Running docker on the branch now. I will give a report at the meeting.

goodb:
Let me know if it doesn’t work.  Maybe it can be fixed before the meeting.  If something is wrong it ought to be a minor fix

ukemi:
It seems to have worked. But there are tons of un-tracked files, including the new imports.

PR #16714, opened 2018-11-29
Update only_in_taxon.tsv

ukemi:


balhoff (commented):
Should it be standard practice to remake the taxon constraints import in the same PR when editing a TSV?

ukemi:
Hmmmm. That is a good point. But I always edit the tsv files like this. If we want to do everything in the same PR (probably best practice), then what is the best route, to edit both files on a local branch?

balhoff:
Oh you edited on GitHub? If you want you can do a `git pull` on master on your local repo, and I think it will bring down all the new branches. Then `git checkout ukemi-patch-2`, run the makefile, then push. Or you could just do your normal workflow and edit the TSV on your desktop.

ukemi:
I just started doing the former. Let's see how that works.

balhoff (approved):

ukemi:
@balhoff, I also updated the documentation wrt the procedure we used here.

Issue #11066, opened 2014-05-27
Change relation for metabolism terms


gocentral:
This is actually multi-step
1. Define new relation in RO (cjm, DH, DOS)
2. clear TG queue and Freeze editing
3. Search and replace in editors file
4. Change TG template
5. Unfreeze

Reported by: cmungall

Original Ticket: [geneontology/ontology-requests/10881](https://sourceforge.net/p/geneontology/ontology-requests/10881)

cmungall:
Fixed with ##16910

PR #16797, opened 2018-12-13
REPLACED by #16886  has-input refactor, Geneva meeting

cmungall:
Dependent on https://github.com/oborel/obo-relations/pull/291

 #11066

cc @ukemi @vanaukenk @pgaudet


ukemi:
Spot-checking this branch

1. metabolic process terms: passes spot checks (logical defs and reasoner checks)
2. biosynthetic process terms: logical defs have not changed. 'has primary output' is missing
3. catabolic process terms: logical defs have not changed. 'has primary input' is missing.
4. 'has primary input or output' is missing a definition (how do we define 'primary'?)
5. 'has primary input or output' is not a child of has_participant
6. Should we obsolete 'results in breakdown of' and 'results in formation of'? The definition of the former is necessary, but not sufficient for what this term's label describes. The latter is undefined.

ukemi:
'has primary input or output'--a relation between a process and a continuant, in which the continuant is somehow involved in the process and is of higher ranked importance than other continuants involved in the process.

ukemi:
Is there a reason why we decided against (perhaps because of the 'results in' terms):

has participant
- has primary participant (instead of 'has primary input or output')
-- has primary input
-- has primary output
-has input
--has primary input
---has ligand (has a domain to an obsolete term)
-has output
--has primary output
--results in formation of (by usage it looks like this is restricted to developmental formation)

deustp01:
Key small molecules and their roles (input, participant, ...) are useful as components of the logical definitions of biological processes, but usage can be tricky. For example, ADP and ATP are certainly key participants in glycolysis but formally assigning this role to them creates the local inference that glycolysis is a kind of nucleotide metabolism.

The list of molecules known to Rhea and the frequency with which each is used, that Anne Morgat and Alan Bridge generated recently, could provide a work-around. No matter how biologically important a molecule is in a process, if it is highly used it is off-limits for logical definitions, while ones used two or three times are candidates for uniquely identifying a process or perhaps linking two closely related processes.

Perhaps a script could take all the molecules associated with a process (from the Rhea reactions linked to the molecular functions that make up the process), find the ones that are least used on the Morgat - Bridge list, and by default make those part of the definition of the process. As resources are available, biologist curators can review those default choices and refine them to make the logical definition fit as closely as possible to the expert consensus view of what the key molecules are.

cmungall:
Replaced by #16886

Issue #15990, opened 2018-07-03
Revise 'GO:0036466 synaptic vesicle recycling via endosome'


BarbaraCzub:
Revise 'GO:0036466 synaptic vesicle recycling via endosome'. Edit definition/modify relations, if required, to fit with the points emphasised by Pim in https://github.com/geneontology/go-ontology/issues/14540

vanaukenk:
@ukemi - should @BarbaraCzub Barbara actually be assigned to this one?
You can give me another ticket in it's place :-)

BarbaraCzub:
Hi @vanaukenk and @ukemi, I opened this issue, because it's been hidden in another ticket, and it temporarily disappeared off the radar. But I need to prioritise annotating, so unless I happen to need this term for annotation, I will not be making any revisions to it in the near future (cc @RLovering). Since this has not been addressed for so long anyway, I assume this is not urgent?

BarbaraCzub:
https://github.com/geneontology/go-ontology/issues/16825

BarbaraCzub:
According to PMID:15217342 'GO:0036466 synaptic vesicle recycling via endosome' refers specifically to vesicles endocytosed via the clathrin-mediated route, and it is the process occurring _after_ clathrin-mediated endocytosis. It occurs in the presynaptic neuron. This is shown in PMID:15217342 in Figure 1 in steps 8 and 9 and in more detail in Figure 3.

Therefore, I will update the current definition only slightly to make it clear that the endosomal recycling occurs _after_ _clathrin-mediated_ endocytosis:

Current:
"Synaptic vesicle recycling where vesicles endocytose via clathrin-coated pits, re-acidify, and refill with neurotransmitters after passing through an endosomal intermediate. PMID:15217342"

Revised:
"Synaptic vesicle recycling where vesicles endocytose**d** via clathrin-coated pits re-acidify and refill with neurotransmitters after passing through an endosomal intermediate. PMID:15217342"

I will also remove the current relationship to 'postsynapse' ('synaptic vesicle recycling' is a part of 'synaptic vesicle cycle', which occurs in the 'presynapse') (please see Figure 1 in PMID:15217342).

At the same time I'll remove the relation between 'synaptic vesicle endosomal processing' and 'synaptic vesicle cycle' because the processing is _not_ a part of the cycle according to the GO term overview (please see screenshots below).

[...]

PR #16921, opened 2019-02-13
Updated definition

BarbaraCzub:
cc @RLovering


BarbaraCzub:
https://github.com/geneontology/go-ontology/issues/15990

PR #16928, opened 2019-02-18
added back taxon contraint on iron import across cell outer membrane

pgaudet:
@ukemi can you please review and merge if you agree? 

Thanks, Pascale


ukemi:
Also regenerated import files and merged. This should all be complete now.

PR #16930, opened 2019-02-19
added do not annotate label

pgaudet:


pgaudet:
Redid correctly here: https://github.com/geneontology/go-ontology/pull/16987

PR #16932, opened 2019-02-20
add oboInOwl to contrib/; work on geneontology/geneontology.github.io #142

kltm:


kltm:
Fix discussed with @cmungall

Issue #12344, opened 2016-03-15
Remove regulates o part of -> regulates property chains


dosumis:
Therese property chains are useful as heuristics as often correct, but are also dangerous. Some processes (e.g. signal transduction) consist of a series of subprocesses/activities that form a causal/regulatory chain. These activities do not regulate the overall process. 

Recent example: inference of reg synaptic vesicle exocytosis - a key step in synaptic transmission - as a modulator of synaptic transmission.

NOTE: removal needs to be done with care given that many inferences are correct.  Need to think through how fits with ETINE review.

(Ticket might be better on RO? Also suspect this is at least partly addressed in other tickets)

ukemi:
I think this was the right move for now. Originally Tanya and I proposed that regulates would be reserved for regulatory activities that completely outside of the process being regulated. It appears now that the pendulum has swung back and even in Noctua models, activities directly activate other activities that are part of the same process. This creates problems with this property chain.

ukemi:
One of the things Kimberly and I have been talking with Heiko about wrt annotations derived from models is the possibility of having the precedence rule that if two terms have both a regulates and an asserted part of relation between them, then the part of relation should take precedence instead of creating an annotation to regulation of the latter term. This would work for annotations derived from models.

cmungall:
@dosumis Before going ahead, can we examine this case in more detail?

This is the hierarchy:

```
     is_a GO:0050804 ! modulation of synaptic transmission
      is_a GO:0046928 ! regulation of neurotransmitter secretion
       is_a GO:2000300 ! regulation of synaptic vesicle exocytosis ***
```

are we sure we have the logical def for GO:0050804 correct? For example, perhaps by 'modulation', we mean something like 'directly regulates'? If we defined it as such, this would block the inference, since the inference comes from the transitivity of regulates.

I suspect transitivity may be real issue here. We wouldn't expect control to be retained over long chains, but we can't really model noise in OWL. My proposed solution is to read the generic regulates in a very weak way, as saying "causally upstream of, and there is a control signal in there, which may influence the target". But this is maybe too unintuitive, and we may want to break transitivity of regulates altogether. We would keep the generic regulates as encompassing direct and indirect cases, but the indirect cases would need to be asserted.

cmungall:
@ukemi - OWL doesn't do precedence, but your rule can be formulated in such a way that we are filtering based on redundancy rather then precedence

ukemi:
[...]

PR #16964, opened 2019-02-27
Strip 'regulates o part_of' chains after reasoning, before releasing ontology.

balhoff:
We will soon be removing 'regulates o part_of -> regulates' chains from the ontology: #12344

Unfortunately I over eagerly [added some rules to RO](https://github.com/oborel/obo-relations/pull/300) which illustrate the problem with these chains; @goodb has pointed out that having both now in Noctua is causing many logical inconsistencies, especially with models imported from Reactome.

Since the regulates chains are on their way out, but waiting for @ukemi to review the existing subsumptions they were providing, I would like to filter them from the release after they have been used to compute the classification. This way the GO hierarchy remains the same but we don't get the consequences of the chains in Noctua models.

This is a little hacky but it's causing an acute disruption in Noctua at the moment.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -183,8 +183,8 @@ enhanced.owl: $(SRC) imports/go_taxon_constraints.owl imports/reactome_xrefs_imp
 
 # reasoned.owl is equivalent to editors source, after reason-relax-reduce pipeline
 # note: we keep this as a distinct intermediate target as the ontology IRI needs to be 'go' for Oort to work...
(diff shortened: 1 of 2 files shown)


ukemi:
Jim,

Another option would be for you to look at the spreadsheet and look for the green line. That’s how far I got. We could instantiate all of the relationships above the green line that don’’t have a yes in the second column. Then you could instantiate all the one below the green line and I will just go through and edit the ontology. It is no worse than we are now because we are inferring all of these anyways.

-D

PR #16984, opened 2019-03-04
added do not annotate tag fixes #16902

### LINKED ISSUE #16902: Add do_not_annotate_directly to multi-organism process
Note we have this E coli gene that is annotated directly to MOP:

http://amigo.geneontology.org/amigo/reference/PMID:23667443

Can we not provide a more specific term in cases like this? cc @jimhu-tamu

pgaudet:


pgaudet:
Redid correctly here: https://github.com/geneontology/go-ontology/pull/16987

PR #17085, opened 2019-03-29
Obsolete some GOREL properties in ro_pending, which are now in RO.

### LINKED ISSUE #16992: RO version of 'results in organization of' *not* available for editing, but GOREL version is
I was creating a logical definition for 'flagellum attachment zone organization' today and didn't get the expected inferred parentage.

@ukemi and I looked at this and realized that in the editing window for creating the logical def, only the GOREL version of 'results in organization of' was available for use in the autocomplete.  The RO version of the relation was not there, even though existing terms in the ontology use it in their logical defs.

Eventually, we worked around it by adding the RO relation using its IRI, but we're unclear as to why the RO version isn't available for editing.

balhoff:
Also updated axioms to use RO 'results in remodeling of' instead of GOREL version. This may result in additional classifications (we need an easy way to report this!).

Fixes #16992.


balhoff:
@pgaudet do we need to announce this kind of obsoletion?

cmungall:
No, groups are using the snakecase labels; double checking with @alexsign

balhoff:
> No, groups are using the snakecase labels

I don't totally follow. I removed a bunch of snakecase labels a while back. These are just a few that had GOREL IRIs.

cmungall:
I mean when groups publish GAF, c16 uses the label in snakecase, so as long
as these are preserved, things should be fine

>

balhoff:
Aha, now I get it, thanks.

PR #17119, opened 2019-04-03
created solute-anion antiporter activity, merged ion antiporter activ…

### LINKED ISSUE #16978: antiporters
_From @dosumis on December 13, 2016 15:59_

_From @dosumis on March 17, 2016 11:20_

I'm sure this issue has been discussed many times before.  Coming up again because it is causing confusion in the synapse project.

In general usage, 'transporter activity' involves active transport:

e.g. see 
https://en.wikipedia.org/wiki/Ion_transporter
https://en.wikipedia.org/wiki/Neurotransmitter_transporter
https://en.wikipedia.org/wiki/Dopamine_transporter

For passive transport (down a concentration gradient), the term 'channel' is used.

These are the uses that electrophysiologists/neurobiologists expect.

But in GO we have

transporter activity (1089)
.  transmembrane transporter activity (938 subclasses)
.  .  active transmembrane transporter activity  (427 subclasses)
.  .  .  active ion transmembrane transporter activity (0 subclasses!)
.  .  passive transmembrane transporter activity (164 subclasses)
.    .  channel activity (162 subclasses)

Issues:
- Given normal usage, I have a strong suspicion that much annotation to the 347 'transmembrane transporter activity' subclasses not under the active or passive branches, are actually active transport.  Is the active/passive distinction pushed down a level because annotators want and MF term even if they don't know whether transport is active or passive?  Wouldn't the process class be enough in that case?  Given this should we consider moving this differentium up?
- Many classes that should be under 'active transmembrane transporter activity' or 'active ion transmembrane transporter activity' are not.

Taking calcium ion transporter as an example:

image

'calcium-transporting ATPase activity' is  not classified under  'active transmembrane transporter':

image

Immediate fixes:
- [x]  'hydrolase activity, acting on acid anhydrides, catalyzing transmembrane movement of substances' _SubClassOf_ 'active transmembrane transporter activity'
- [x] 'ATPase coupled ion transmembrane transporter activity' _SubClassOf_ 'active ion transmembrane transporter activity'

' calcium:cation antiporter activity' is not classified under 'active ion transmembrane transporter activity'

image

https://en.wikipedia.org/wiki/Antiporter

antiporter activity: "Enables the _active_ transport of a solute across a membrane by a mechanism whereby two or more species are transported in opposite directions in a tightly coupled process not directly linked to a form of energy other than chemiosmotic energy. The reaction is: solute A(out) + solute B(in) = solute A(in) + solute B(out)."

Ditto symporter.

Fix: 
- [x]  add 'ion antiporter activity' _SubClassOf_ 'active ion transmembrane transport activity'
- [x] add 'anion:cation symporter activity' _SubClassOf_ 'active ion transmembrane transport activity'

But - in this case we may need to watch out for incorrect inferences that may result, given that only one of the transported ions is actively transported in these cases.

Relevant historical discussion: http://gocwiki.geneontology.org/index.php/Tranporter_Activity_Meeting_Notes

_Copied from original issue: geneontology/go-ontology#12347_

_Copied from original issue: geneontology/molecular_function_refactoring#26_

pgaudet:
…ity into antiporter activity, fixes #16978


cmungall:
I'm concerned about the loss of definitional axioms

pgaudet:
What do you mean ?

Issue #16547, opened 2018-10-31
Does synaptic vesicle transport occur in the cytoplasm?


cmungall:
This is an inference we get

See PR #16546

cc @lpalbou

ukemi:
The structure is not in alignment with exocytosis. Seems incorrect to say that exocytosis occurs in the cytoplasm, but 'synaptic vesicle exocytosis' is a child of 'synaptic vesicle transport'. Isn't exocytosis more than just transport of the vesicle? I would be ok with it being a 'vesicle-mediated transport' because that refers to the contents as well. I think we need to clean up this are of the ontolgy.

ukemi:
Is a cytoplasmic vesicle part of the cytoplasm? Or is it located in but distinct from the cytoplasm?

pgaudet:
I was wondering if this is strictly an intracellular process - for example synaptic vesicles are endocytosed, so they exist extracellularly for a time ? 

https://en.wikipedia.org/wiki/Synaptic_vesicle#The_synaptic_vesicle_cycle

ukemi:
OK. It fits the definition then. So if a synaptic vesicle is always part of the cytoplasm, then it's transport necessarily occurs in the cytoplasm. I'm not sure the property chain is the problem here.

ukemi:
But I don't think the vesicles are endocytosed. I think that the neurotransmitter in the synaptic cleft is endocytosed to form new synaptic vesicles. Clearly we need an expert to comment on this.

pgaudet:
I dont know either whether vesicles are endocytosed - however wikipedia states "6. Endocytosis

This accounts for the re-uptake of synaptic vesicles in the full contact fusion model. However, other studies have been compiling evidence suggesting that this type of fusion and endocytosis is not always the case. "

BarbaraCzub:
This ticket is possibly related: https://github.com/geneontology/go-ontology/issues/15989

BarbaraCzub:
https://github.com/geneontology/go-ontology/issues/15990

BarbaraCzub:
According to the Sudhof review of synaptic vesicle cycle (PMID:15217342, Figure 1 and Figure 3), synaptic vesicles always remain inside the cell. Therefore, I think it is correct that they are always a part of the cytoplasm.

[...]

PR #17200, opened 2019-04-29
Removed link between synaptic vesicle transport and vesicle-mediated …

BarbaraCzub:
…transport in synapse #16547

Hi @ukemi 
When I removed the link between 'synaptic vesicle transport' and 'vesicle-mediated transport', the reasoner inferred the parent of  'synaptic vesicle transport' as 'synaptic vesicle localization'. However, previously the parent asserted manually was the more specific 'establishment of synaptic vesicle localization' (screenshot below). Would you agree I should remove 'establishment of synaptic vesicle localization' if 'synaptic vesicle localization' if inferred by the reasoner? I would also remove the manually asserted relation to 'transport'. Could you please confirm whether I should go ahead?

Re1

Thanks,
Barbara

cc @RLovering


ukemi (commented):
I think this is the appropriate course of action. I suspect it is due to some discussion on transport and the process of localization versus the establishment of the localization. For synaptic vesicles, there must have been a conclusion that the transport was the localization process. You should remove the asserted parents, rerun the reasoner and make sure the inferred superclasses are still correct.

BarbaraCzub:
Ok. I will do. Thanks you.

BarbaraCzub:
Looks fine:
Re1b

PR #17208, opened 2019-04-30
refined relations between endocytic process terms and added PMIDs whe…

BarbaraCzub:
…re they were missing #16547

cc @RLovering


ukemi (commented):
Hi,

Have another look at GO:0098881. I don't think there should be two genuses in the logical definition. Seems like there is an error.

BarbaraCzub:
Thanks @ukemi 
I included both terms in the logical definition, because the process of 'exocytic insertion of neurotransmitter receptor to plasma membrane' involves both 'vesicle docking' and 'vesicle tethering'. But perhaps I should remove the logical definition all together then, and manually assert the relationships with these parent terms?
Thanks,
Barbara

ukemi:
Yes. That makes more sense to me.

BarbaraCzub:
Ok. Thanks. I'll remove the logical definitions and assert parents manually.

PR #17231, opened 2019-05-02
changed logical def xenobiotic transport fixes #17226

### LINKED ISSUE #17226: Parentage of  GO:1990961 xenobiotic detoxification by transmembrane export across the plasma membrane
Shouldn't this term ( GO:1990961 xenobiotic detoxification by transmembrane export across the plasma membrane) somehow be related to GO:0042908
xenobiotic transport? 

From ticket https://github.com/geneontology/go-annotation/issues/2332

pgaudet:


balhoff:
@pgaudet I came across this while working on removal of 'substance with role' logical definitions. I think this change should have gone the other way. Instead of changing 'xenobiotic transport' to refer to 'xenobiotic', it should continue to refer to 'substance with xenobiotic role'.

'xenobiotic detoxification by transmembrane export across the plasma membrane' should also refer to 'substance with xenobiotic role' and not 'xenobiotic'.

ukemi:
I think we had the same issue with estrogen. Any way we can check to be sure we haven't directly used roles in logical defs?

balhoff:
I fixed estrogen also. I did a DL query for ‘has participant’ some role. Actually we could make that expression equivalent to Nothing to stop it happening.

ukemi:
We should also make the direct roles unavailable in Noctua.

pgaudet:
Did you fix it ? Let me know if I need to do anything.

balhoff:
@pgaudet I fixed in #19528

Issue #17435, opened 2019-06-01
move MetaCyc:CODH-PWY from acetate biosynthetic process from carbon monoxide to carbon fixation by acetyl-CoA pathway


cmungall:
https://biocyc.org/META/NEW-IMAGE?type=PATHWAY&object=CODH-PWY

![image](https://user-images.githubusercontent.com/50745/58754453-06233c00-8485-11e9-8043-f2b91cf0c860.png)

```
id: GO:0019415
name: acetate biosynthetic process from carbon monoxide
namespace: biological_process
def: "The chemical reactions and pathways resulting in the formation of acetate from other compounds, including carbon monoxide." [GOC:go_curators]
synonym: "acetate anabolism from carbon monoxide" EXACT []
synonym: "acetate formation from carbon monoxide" EXACT []
synonym: "acetate synthesis from carbon monoxide" EXACT []
synonym: "carbon monoxide dehydrogenase pathway" EXACT []
xref: MetaCyc:CODH-PWY
intersection_of: GO:0009058 ! biosynthetic process
intersection_of: has_primary_input CHEBI:17245 ! carbon monoxide
intersection_of: has_primary_output CHEBI:30089 ! acetate
```

should be here:

```
id: GO:0030634
name: carbon fixation by acetyl-CoA pathway
namespace: biological_process
def: "A pathway of carbon dioxide fixation in which one molecule of acetyl-CoA is completely synthesized from two molecules of carbon dioxide (CO2)." [PMID:11607093]
synonym: "Ljungdahl-Wood pathway" EXACT []
xref: MIPS_funcat:01.05.01.05.07
xref: Wikipedia:Wood–Ljungdahl_pathway
is_a: GO:0006085 ! acetyl-CoA biosynthetic process
is_a: GO:0015977 ! carbon fixation
```

cmungall:
I guess this is not totally straightforward. The LW pathway is sometimes extended to acetate as in here:

![image](https://user-images.githubusercontent.com/50745/58754623-4e445d80-8489-11e9-88ef-326b443b41b7.png)

here GO:0030634 would be a part of the LW pathway (see above)

in other cases it stops at acetyl-CoA (MetaCyc has the acetyle-CoA to acetate part as a separate pathway). Here GO:0030634 would *overlap* LW/

The fact that different pathway databases consider different start/endpoints is well known, but we should pick one and be very clear on our starts/ends.

I think GO:0019415 may be a candidate for obsoletion

[...]

PR #17436, opened 2019-06-02
Fixes for Ljungdahl-Wood pathway #17435

cmungall:
Adding taxon constraint, never in Eukaryotes
(we need a way to give a PMID for TCs!)

Added synonyms with provenance

Moved MetaCyc xref, see #17435

(but we may still want to obsolete the CO form, see ticket)


ukemi (commented):
Looks like you have a bad character in the Wikipedia xref.

ukemi:
@cmungall, do you want me to go ahead and finish this up for you? It looks like we still need to run docker and generate the taxon owl file.

PR #17529, opened 2019-06-24
making the release date injectable for better control during release

dougli1sqrd:
This makes the `RELEASEDATE` picked up by the environment if it's there using the `?=` operator in Make.


kltm:
@dougli1sqrd This make sense to my eyeballs. I'd like @cmungall to chime on if possible, noting that @dougli1sqrd is creating a minimum change according to what I asked.

kltm (approved):

cmungall (approved):

PR #17545, opened 2019-06-28
Slightly faster multiple-class-links query.

balhoff:


balhoff:
Closing since I don't think this makes as much of a difference as I thought.

Issue #17548, opened 2019-06-28
obsolete term GO:0104005 used in taxon constraint


balhoff:
@ukemi could you check on what to do with this? I am enhancing the makefile to catch these, so they shouldn't be lurking in the future.

ukemi:
@balhoff, part of the editors' standard procedure when obsoleting a term should be to check the taxon constraints and either remove them or change them as appropriate. Of course this gets a bit clunky since we will have to manually rebuild the taxon import before we can obsolete the term. We really need to streamline the import generation procedure in some way.

ukemi:
It looks like this term was obsoleted with no replacement. It should just be removed from the constraints.

ukemi:
Removed obsolete term and tested with ontobot.

PR #17549, opened 2019-06-28
Expand check for obsolete usages to be much more comprehensive.

balhoff:
Restrict checks to GO and not imports.


balhoff:
I added a target which can be used to look for usages of obsoletes within imports.

balhoff:
This will not pass until #17548 is fixed.

PR #17566, opened 2019-07-01
Revert "Update CL import." 

balhoff:
This reverts commit 135314d7c79e6495caf9f284a975999b715a1787.

There are some weird issues with missing cell labels. cc @ukemi


balhoff:
It turns out cl-basic has lost all its labels and definitions.

PR #17606, opened 2019-07-09
possible fix for geneontology/go-site#1139

kltm:
For https://github.com/geneontology/go-site/issues/1139


goodb (approved):
I think this improves things.  Build working, new file based on new KEGG_Reaction namespace makes sense.

PR #17615, opened 2019-07-12
fixed EC number

pgaudet:


kltm:
https://github.com/geneontology/helpdesk/issues/224#issuecomment-510932044

PR #17619, opened 2019-07-15
fixed ec for helpdesk ticket

pgaudet:


kltm:
@pgaudet As a note, if you add the tag the other way, e.g.:
https://github.com/geneontology/helpdesk/issues/224
will show the status as well in the other ticket.

PR #17624, opened 2019-07-16
Issue 17623

pgaudet:


ukemi (commented):
I scrolled through these and they all looked good to me. It seems like we never changes the definitions on the legacy obsoletions. I can't figure out why there were two commits, but they both look ok to me. Nice catch with the synonyms.

pgaudet:
@ukemi or @cmungall Can you have a look at the PR and let me know if this is OK ?

PR #17695, opened 2019-08-08
[DO NOT MERGE] Declare GO cellular_component to be a subclass of CARO anatomical entity.

balhoff:
Required for GO-CAM schema checking.


ukemi (approved):

balhoff:
Before merging this we need to decide on whether it is okay for complexes to be anatomical entities.

ukemi:
In FMA, even macromolecules are anatomical strucutures. As @vanaukenk pointed out, in CARO they appear to be restricted to a granularity above anatomical structure.

Issue #17506, opened 2019-06-17
Fix KEGG Reaction links


pgaudet:
From https://github.com/geneontology/amigo/issues/568

In fact the prefixes are incorrect in the ontology: 
xref: KEGG:R#####
should be xref: KEGG_REACTION:R#####

pgaudet:
Updated Makefile #19023

PR #17731, opened 2019-08-19
Do not produce kegg2go because now we use kegg_reaction in the ontology

cmungall:
https://github.com/geneontology/go-ontology/issues/17506


pgaudet:
@cmungall can this be merged ?

PR #17799, opened 2019-09-06
Fix part term children with missing parent terms

balhoff:
@ukemi please take a look. I will obsolete 'intracellular organelle part' and 'growth cone part' in a separate PR, since I guess we will need to announce those. I temporarily put them back under cellular_component here.


ukemi (approved):
Looks good. From the diff it looks like we covered everything. We should open an issue to describe this just for tracking purposes. We can close it once we are sure we cleaned this up. Thanks for taking care of this @balhoff

PR #17800, opened 2019-09-06
More part obsoletions

balhoff:
'intracellular organelle part' and 'growth cone part'

(diff will be smaller once #17799 is merged)


pgaudet:
These terms don't have annotations. You can merge this. I'll add them to the announcement. 

Thanks, 
Pascale

Issue #17822, opened 2019-09-12
fix pyruvate dehydrogenase activities


rwst:
```
EC:1.2.1.51 > GO:pyruvate dehydrogenase [NAD(P)+] activity ; GO:0034603
EC:1.2.1.51 > GO:pyruvate dehydrogenase (NADP+) activity ; GO:0050243
```
enzyme.dat:
```
ID   1.2.1.51
DE   Pyruvate dehydrogenase (NADP(+)).
AN   Pyruvate:NADP(+) oxidoreductase.
CA   Pyruvate + CoA + NADP(+) = acetyl-CoA + CO(2) + NADPH.
```
So, EC:1.2.1.51 === GO:0050243, and GO:0050243 is_a GO:0034603. Thus, GO:0034603 is a superset of EC:1.2.1.51 and so can no longer have an association with it.

I'll submit a PR.

rwst:
@pgaudet Is the merge final, so this ticket and the PR can be closed?

Re: NAD(P): there are 159 definitions containing the string

pgaudet:
Ah ! This is not a `git merge` - I was suggesting to merge the 2 terms. 

I think @hdrabkin should have a look at these before we finalize. He's off until Monday, it should be done early next week. 

Thanks, Pascale

deustp01:
@pgaudet said 

> @ukemi Did we not say we would get rid of the "NAD(P) " terms any only keep the specific ones ?

We may have (forgetfulness here) but, there is maybe one mammalian enzyme that can use either  NAD or NADP, and probably in other non-vertebrate domains of life that don't use the mammalian strategy to maintain separate pools of reducing equivalents for energy generation (NAD) and for biosyntheses (NADP) there are more.

My recollection of @ukemi 's fix was that it was directed more to sorting out parentage in the ontology, to properly show relationships among functions that require NAD, ones that require NADP, and ones where the specific requirement NAD vs NADP has not been established? So sorting @rwst 's 159 definitions into these three categories looks like a next step.

pgaudet:
OK, thanks for the reminder.

ukemi:
@deustp01, that is correct. The discussion was about NAD(P) in the definitions and whether or not those classes represented both reactions or either reaction. From an ontological standpoint since ChEBI has NAD and NADP as children of NAD(P), we decided it had to mean either. If we know that a particular 'gene product' can catalyze both, we would annotate to both children.

pgaudet:
[...]

PR #17824, opened 2019-09-12
fix pyruvate dehydrogenase activities #17822

rwst:


rwst:
Probably irrelevant now.

Issue #17696, opened 2019-08-08
NTR: cellular anatomical entity


balhoff:
For the GO-CAM schema, we would like to group cellular components under CARO anatomical entity, so that we can state that a molecular function instance occurs in 0 or 1 anatomical entity (being a cellular component, cell, or gross anatomical structure). I opened a PR #17695 to this end, but it introduces a problem in that it makes protein-containing complex a subclass of anatomical entity. CARO anatomical entity states it has "granularity above the level of a protein complex". Complexes are a bit of a problem as anatomical entities anyway, because (I think) we don't want to have GO-CAMs stating that functions occur in complexes.

There is currently no term that groups other cellular components to the exclusion of protein-containing complex. Could we add a term like 'cellular anatomical entity' as a child of 'cellular_component', with all cellular components as children with the exception of 'protein-containing complex'? Then in `go-bridge.owl` we would make 'cellular anatomical entity' a child of CARO 'anatomical entity'.

pgaudet:
Interesting. I had understood "granularity above the level of a protein complex" to exclude protein complex (as in >, not >=). 

Pascale

cmungall:
This parallels exactly a conversation we are having in obo core, I will link to here when we have a ticket

ukemi:
It seems this will be needed if we want to successfully coordinate with CARO and intend to use their definition of anatomical entity.

balhoff:
@ukemi or @vanaukenk would you mind creating a pull request for this? I am not set up for minting term IDs.

vanaukenk:
Will do.

ukemi:
https://github.com/geneontology/go-ontology/pull/17704

ukemi:
Looks like we created the term and merged, but we still haven't moved the "xyz part' terms to be direct children of cellular component. We need to do this and then assert everything under those that is not a protein-containing complex to be a cellular anatomical entity. Otherwise some protein-containing complexes are asserted or inferred to be cellular anatomical entities. For example look at 6-phosphofructokinase complex.

ukemi:
Should we unmerge #17704 or just work from it as a starting point?

pgaudet:
Ah ! Perhps we should undo the changes, then. 

@balhoff Would you be doing the script to move the part terms under the new term ? Any idea how long that would take to do ? 

Thanks, Pascale

ukemi:
[...]

PR #17946, opened 2019-10-03
Make 'cellular anatomical entity' subclass of CARO 'anatomical entity'; disjoint with 'protein-containing complex'

balhoff:


balhoff:
See #17696.

Issue #17625, opened 2019-07-16
Add Travis check for obsolete GO term definitions


pgaudet:
Hello, 

The definitions of obsolete GO terms should begin with "OBSOLETE. " This was not the case (see #17623). Can we add a Travis check for this ? 

Thanks, Pascale

ukemi:
Not sure who is writing these now, but have tagged who I think are the appropriate people.

balhoff:
@dougli1sqrd implemented this.

PR #17955, opened 2019-10-03
adding sparql check and adding the run to the makefile

dougli1sqrd:
For https://github.com/geneontology/go-ontology/issues/17625

### DIFF
--- a/src/sparql/obsolete-definition-violation.sparql
+++ b/src/sparql/obsolete-definition-violation.sparql
@@ -1,0 +1,20 @@
+# This checks if a class definition does not begin with "OBSOLETE"
+# in the definition (http://purl.obolibrary.org/obo/IAO_0000115) of the term
+
+PREFIX definition: <http://purl.obolibrary.org/obo/IAO_0000115>
+prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+prefix oio: <http://www.geneontology.org/formats/oboInOwl#>
+prefix owl: <http://www.w3.org/2002/07/owl#>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix xsd: <http://www.w3.org/2001/XMLSchema#>
+
+SELECT ?term ?definition
+WHERE {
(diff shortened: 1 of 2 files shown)


src/sparql/obsolete-definition-violation.sparql
```diff
@@ -0,0 +1,20 @@
...
+prefix xsd: <http://www.w3.org/2001/XMLSchema#>
+
+SELECT ?term ?definition
+WHERE {
+    # Find deprecated terms (obsolete)
+    ?term   a   owl:Class ;
+            owl:deprecated  "true"^^xsd:boolean ;
+            definition: ?definition .
+    
+    FILTER(!strstarts(?definition, "OBSOLETE"))
```
balhoff: Looks good. I think you can make this stronger and check for "OBSOLETE. " (period and space)

balhoff (approved):

src/sparql/obsolete-definition-violation.sparql
dougli1sqrd: oh yes? I didn't realize that was part of the standard. I'll add it.

dougli1sqrd: Are you sure we need the space after the period? There are a handful of terms that have no space:
`http://purl.obolibrary.org/obo/GO_0006981,OBSOLETE.The conversion of the SoxR transcription factor to its active (oxidized) form.`

Many other terms have no other definition other than: `OBSOLETE.` like for `http://purl.obolibrary.org/obo/GO_0015096` (which also does not contain a space).

balhoff: In my opinion the terms without spaces are mistakes. Not sure about a definition that is only "OBSOLETE.". That seems like a mistake as well, but in the example you give, there is a definition xref attached.

dougli1sqrd: Ah I see. You think there should be a real definition there too? Should this rule be the one to catch that also? I guess it can. I'll keep the space in.

dougli1sqrd:
Thanks for the update @balhoff

PR #17964, opened 2019-10-05
Stick with cl-basic for now, until incompatibilities are addressed.

balhoff:
See https://github.com/obophenotype/cell-ontology/issues/584


kltm:
Tagging @kltm

Issue #18009, opened 2019-10-15
Logical definition for sperm motility is missing


ukemi:
Need to add logical def.

krchristie:
I know it looks weird that the term `sperm motility` doesn't have a logical def and the term `flagellated sperm motility` does have one that invokes a term called just `sperm` (CL:0000019 ! sperm), so I want to remind you of the history from this ticket:

https://github.com/geneontology/go-ontology/issues/15945 
> The term for 'sperm motility' was created by Paola and myself when we also created the term for 'amoeboid sperm motility' term to make the distinction between that and 'flagellated sperm motility'. If I recall correctly, the existing term for flagellated sperm motility was originally named just 'sperm motility', but was defined such that it was clearly flagellum-dependent, so we opted to keep it as the specific term, consistent with its definition.
> 
> My recollection of the logical def is that the existing CL term for 'sperm' specifies a flagellated sperm, so it seemed sufficient for the logical definition of 'flagellated sperm motility'. I think we discussed whether the CL term for sperm should be changed, but I have vague recollections that we were told CL focuses on mammalian cells, so we didn't attempt to get that changed.
> 
> Here is the CL term for 'sperm':
> http://bioportal.bioontology.org/ontologies/CL?p=classes&conceptid=CL%3A0000019

At the time, we created the term `sperm motility`, Paola and I were not able to figure out how we could make logical definitions for either `sperm motility` or `amoeboid sperm motility` when the CL term is specific to a flagellated sperm so weren't aware of terms we could use to refer to a sperm generally or to an amoeboid sperm.

cmungall:
CL is broader than mammals, but you are right to be cautious about making
assumptions about the exact meaning of classes if it's unclear.

CL is a bit squishy as to whether it pertains to metazoa or broader. We
need to have more TCs in CL so people are not guessing.

I assume GO sperm motility also needs to apply to plants (again: on our
side, we could be more explicit. We have a lack of TCs here so the default
assumption is broadly applicable, but it's always good to be explicit, e.g.
include examples in the definition gloss). So it is unsafe to use the CL
class if it is cryptically more restricted.

Here is how I would suggest approaching issues like this: Just as GO exists
to serve a community, so does the rest of OBO, and that includes serving
GO. If OBO isn't doing a good job in providing the concept GO needs then I
recommend being more pushy. Furthermore, we should be trying to use
biological language consistently, so if GO uses "sperm cell" differently
[...]

PR #18121, opened 2019-11-05
Added logical def for cilium movement and fixed up a few other terms.…

### LINKED ISSUE #18112: Logically define 'cilium movement' as 'movement of cell or subcellular component' and 'results in movement of' some 'cilium'?
To capture `movement of cell or subcellular component`, in particular, GO:0003341 - cilium movement logically, we would like to 

1) know which RO relationship is appropriate to capture the relation between the cell movement and the cilium. @dosumis has suggested RO:0002565 ("results in movement of"), e.g. GO:0060285 ---[has active participant]-->"cilium"
2) assert, in particular, the fact that GO:0003341 equivalentTo: GO:0006928  and `results in movement of` some GO:0005929

EDIT: replaced proposal from `has active participant` to `results in movement of`

ukemi:
… Fixes #18112


raymond91125:
Fixes #18009

PR #18191, opened 2019-11-18
Link DDANAT cell to CL ‘native cell’.

balhoff:
Only added one axiom—some of the other axioms got shuffled around.


vanaukenk:
Looks good, @balhoff 

Thank you!

PR #18242, opened 2019-11-21
fix to issue 18165

hdrabkin:


hdrabkin:
I have no idea why the typedefs moved because I didn't touch them; they seem to be all ROs. This ticket was an NTR and an additional parent change for a related term per ticket 18165

Issue #18362, opened 2019-12-04
add check for namespace validity


balhoff:
See #18356.

ukemi:
@balhoff, I know this check used to be in place because I have had it fail in the past. In fact, I remember you helping me trace down the error in the Travis output.

pgaudet:
I know it was picking up when it was completely missing - was it also picking up typos ? (I also thought so....)

ukemi:
Yup. I had put them in without the underscore by hand and it caught it.

cmungall:
I did a test where I replicate: #18493

it failed as expected

https://travis-ci.org/geneontology/go-ontology/builds/633390910#L235-L236

```
../util/check-obo-for-standard-release.pl --disable-isa-incomplete --xref-abbs GO.xrf_abbs go-edit.obo > go-edit.obo-check.tmp && mv go-edit.obo-check.tmp go-edit.obo-check
 :: ERROR: namespace-different-from-is_a-parent: GO:0150137 in biologica_process, GO:0042089 in biological_process
```

cmungall:
My test did not properly replicate.

This was the original: #18323 

It passed travis.

I have discovered the reason. The check only looks at asserted is-a (subClassOf), not logical defs

ukemi:
Nice detective work @cmungall? @dougli1sqrd, can we modify the test to include both?

Issue #18356, opened 2019-12-04
errors detected in pipeline


balhoff:
It looks like a couple of terms have a misspelled namespace: `biologica_process`.

From @kltm:

```
01:12:54   :: ERROR: namespace-different-from-is_a-parent: GO:0150134 in biological_process, GO:0150133 in biologica_process
01:12:54    :: ERROR: namespace-different-from-is_a-parent: GO:0150135 in biological_process, GO:0150133 in biologica_process
01:12:54    :: ERROR: namespace-different-from-is_a-parent: GO:0150133 in biologica_process, GO:0032673 in biological_process
01:12:54    :: ERROR: namespace-different-from-is_a-parent: GO:0150133 in biologica_process, GO:0050707 in biological_process
01:12:56   
```

kltm:
Can some simple checks be added to catch this type of error?

ukemi:
Weird! I thought these failed the Travis checks. They always have before.

pgaudet:
I thought we used to have a Travis error for this !

ukemi:
I only found one namespace error. But it is fixed and merged.

PR #18493, opened 2020-01-06
DO NOT MERGE deliberate error to investigate #18356

cmungall:
See #18362


ukemi:
Shouldn't we close this PR as well?

PR #18876, opened 2020-02-17
changed parent fixes #18830

### LINKED ISSUE #18830: mating plug formation should not be under behavior
```
id: GO:0042628
name: mating plug formation
namespace: biological_process
def: "The deposition of a plug of sperm or other gelatinous material into the opening of the vulva by a male at the termination of copulation. Probably acts to prevent subsequent matings by other males." [GOC:jl, http://www.wildcru.org/glossary/glossary.htm, http://www.wormatlas.org/glossaries/cglossary.htm, PMID:11267893]
synonym: "copulatory plug biosynthesis" EXACT []
synonym: "copulatory plug deposition" EXACT []
synonym: "copulatory plug formation" EXACT []
synonym: "mating plug deposition" EXACT []
is_a: GO:0019098 ! reproductive behavior
is_a: GO:0045297 ! post-mating behavior
relationship: part_of GO:0007320 ! insemination
```

We have 3 experimental annotations:
http://amigo.geneontology.org/amigo/term/GO:0042628

image

On glancing at the abstracts it seems that the genes involved have a role in seminal viscosity. It seems a stretch to says these genes are involved in behavior in any meaningful way.

There are also two TASs from FlyBase from one paper

Again the paper seems to be about seminal fluid composition
http://amigo.geneontology.org/amigo/reference/PMID:11869341

Ignoring the genes, with a broad term like MPF we can imagine some species where this is a behavioral act distinct from copulation - I don't know any. But this doesn't seem to be the case for annotated species.

I think the marked axioms should be removed:

```
*is_a: GO:0019098 ! reproductive behavior
*is_a: GO:0045297 ! post-mating behavior
relationship: part_of GO:0007320 ! insemination
```

I think this could be made to follow anatomical structure formation DP. mating plug is in uberon.

Some additional notes on the definition:

```
def: "The deposition of a plug of sperm or other gelatinous material into the opening of the vulva by a male at the termination of copulation. Probably acts to prevent subsequent matings by other males." [GOC:jl, http://www.wildcru.org/glossary/glossary.htm, http://www.wormatlas.org/glossaries/cglossary.htm, PMID:11267893]
```

 - the wildcru url is 404 and doesn't seem like an appropriate source for a GO definition
 - the wormatlas URL is wrong it should be https://www.wormatlas.org/glossary/cglossary.htm
 - it is curious wormatlas is referenced yet we have no worm annotations
 - I don't think vulval opening is the correct location for many species
 - the definition says nothing about this being behavioral so not clear why we have logical axioms saying it is. It really helps to keep text definitions and logical axioms in sync!
 - the genus 'deposition' seems OK if you are looking at this from an organismal biology lens but really the genes are not involved in deposition in any normal sense of the word.

So I think not only is behavior the wrong lens, I don't think organismal biology is the lens either. It might be better to reframe this in terms of what the genes actually do - e.g. regulate qualities of the portion of sperm that cause it to stick around more (hypothetically as part of sperm competition)

(I am looking at some of the GO behavior terms as part of https://github.com/obo-behavior/behavior-ontology/issues/101)

pgaudet:


cmungall (commented):
Looks OK, though this leaves poor post-mating behavior with one child... is it really a useful grouping?

pgaudet:
I'll propose to obsolete poor post-mating behavior.

cmungall:
Just noting here that 'oviposition' is a very broad grouping, especially for a behavioral process, across fish, chicken, Dmel, Celegans, ...

pgaudet:
@cmungall what do you suggest to do about the 'broad grouping' of ovoposition ? Should I create a new ticket for that ?

PR #18990, opened 2020-02-27
Delete

pgaudet:
Redundant with https://github.com/geneontology/go-ontology/blob/master/README.md


cmungall:
has a conflict.

I recommend keeping this file as it is convention, but having a single line that points to the wiki

PR #18995, opened 2020-02-27
trying make all in travis

dougli1sqrd:


balhoff:
@dougli1sqrd can we close this? I don't think this target is going to work on travis.

dougli1sqrd:
This was just trying to see if we could make the whole ontology in travis so that we did all reasoning and such. To prevent cases where travis passes but something goes wrong with the whole build in the pipeline.

balhoff:
This would be really good, but I don't think it will be possible with the free travis.

PR #19008, opened 2020-02-28
Update README.md

pgaudet:


cmungall (commented):
This looks good. I'm worried about this getting out of sync though... also with the main website. really all metadata should follow the file around. but go ahead and merge for now.

PR #19016, opened 2020-03-02
Add new term properties ‘editor note’ and ‘term tracker item’.

balhoff:
@cmungall this should fail due to the perl check not liking github URI values for relations.


cmungall:
Let's model as a literal

```
property_value: term_tracker_item https://github.com/geneontology/go-ontology/issues/17407 xsd:anyURI
```

I think this is better formally as it doesn't force any ontological commitment about what the URI denotes. It is also preferable pragmatically it uses a different obo format tag that won't be confused with 'true' relationships. It will also avoid issues in future where we might want to enforce all referenced entities to have labels, types, etc.

cmungall:
we also need to relax non-IRI-value-violation.sparql

balhoff:
> Let's model as a literal

I see, I didn't like this idea at first but now I agree. I'll change the check.

PR #19045, opened 2020-03-07
Reduce named only in go-lego

balhoff:


balhoff:
@cmungall will only reducing named classes for go-lego have a negative impact on how it would be used in Amigo? This just occurred to me—previously I was thinking that reducing property relations was mainly useful for consumers of go.obo/go.owl.

Perhaps this issue will go away anyway when we trim down the neo.owl that is imported into go-lego.

PR #19158, opened 2020-03-25
Pipeline issue 35 neo queries

dougli1sqrd:
For https://github.com/geneontology/pipeline/issues/35#issuecomment-601410116

This adds a make target to run `robot profile`. There's a set of queries at src/sparql/neo along with the profile text file, listing what queries to run. This will produce a report at src/ontology/reports/neo-violations.report. 

This command exits with `1` if it fails.


kltm (approved):
Assuming that you're sure your target is clear, I see no reason not to merge.

PR #19184, opened 2020-03-28
Consolidate travis check into one robot invocation for better performance.

balhoff:


balhoff:
@cmungall @dougli1sqrd I am doing some experiments here to reduce the time required for Travis checks (now 10 min 39 sec vs. previous 22 min 36 sec). The main change is that I have consolidated 3 targets into one invocation of `robot`. This saves a lot of time (ontology is parsed once). I also am running `robot` verify once instead of twice (putting all queries into one list). Converting the ontology to Jena triples inside robot is a major time sink, so better to do it only once (I am also going to look into making this faster in the robot code). Most of the SPARQL checks take at most a few seconds, but one takes ~90 seconds (`replacedby-obsolete-violation`). I haven't found a way to speed this query up.

This target checks all the same stuff as before, but will need to be kept in sync with the more separate targets. I think it makes sense to customize an ideal target for Travis like this.

src/ontology/Makefile
```diff
@@ -496,13 +501,10 @@ VCHECKS = equivalent-classes trailing-whitespace owldef-self-reference synonym-l
...
 VARGS = $(foreach V,$(VCHECKS), $(SPARQLDIR)/$V-violation.sparql)
-sparql_test: $(SRC) reasoned.owl enhanced.owl
+sparql_test: $(SRC) reasoned.owl
 	$(ROBOT) verify --input $< --queries $(VARGS) -O reports/
 	# These purposely aren't checking problems in imports
-	$(ROBOT) verify --input reasoned.owl --queries $(SPARQLDIR)/multiple-class-links-violation.sparql -O reports/
-	# This could be merged into previous invocation after robot bug #518 is fixed
-	$(ROBOT) verify --input enhanced.owl --queries $(SPARQLDIR)/obsolete-reference-violation.sparql -O reports/
-	$(ROBOT) verify --input reasoned.owl --queries ../sparql/missing_superclass-violation.sparql -O reports/
+	$(ROBOT) verify --input reasoned.owl --queries $(SPARQLDIR)/multiple-class-links-violation.sparql $(SPARQLDIR)/obsolete-reference-violation.sparql $(SPARQLDIR)/missing_superclass-violation.sparql -O reports/
```
dougli1sqrd: I didn't know/forgot that robot verify could take more than one query!

dougli1sqrd (approved):

.travis.yml
```diff
@@ -14,4 +14,4 @@ before_script:
   - cd src/ontology
 
 script:
-  - ./run.sh make ROBOT_ENV='ROBOT_JAVA_ARGS=-Xmx6G' travis_test
+  - ./run.sh make -j 4 ROBOT_ENV='ROBOT_JAVA_ARGS=-Xmx6G' travis_test
```
dougli1sqrd: does travis have more than one core for our usage?

dougli1sqrd:
Doing more than one query can be done faster if we use the --tdb option in robot! I've been using it a little and it's nice!

.travis.yml
balhoff: Apparently 2.

dougli1sqrd: does doing more than 2 give us any bang then?

balhoff: I think sometimes, if something is blocked on IO.

cmungall:
> Most of the SPARQL checks take at most a few seconds, but one takes ~90 seconds (replacedby-obsolete-violation).

it seems the underlying OWL representation is a problem, due to the fact the obo2owl uses the CURIE rather than URI. Perhaps having a generic CURIE to OWL translation (e.g. separate robot command, implemented procedurally rather than in SPARL) is the fix here. Or changing obo2owl...

wdduncan:
This suggestion will probably be laughed at ... but I'll suggest in anyway :)
If we want to stick with a robot-centric work flow, could we extend robot to have a set of parsing tools, such as robot.grep, robot.replace, etc. Basically, import functionality of obo-scripts into a sub-library of robot.

PR #17673, opened 2019-07-31
DO NOT MERGE-Update ChEBI import

ontobot:
I've updated the ChEBI import. Here's a diff of how this impacts the classified ontology:

# Ontology comparison

## Old
- Ontology IRI: `http://purl.obolibrary.org/obo/go.owl`
- Version IRI: `http://purl.obolibrary.org/obo/go/releases/2019-07-31/go.owl`
- Loaded from: `file:/projects/stars/jenkins-c8/workspace/GO/update-chebi-import/go-ontology/src/ontology/reasoned-prev.owl`

## New
- Ontology IRI: `http://purl.obolibrary.org/obo/go.owl`
- Version IRI: `http://purl.obolibrary.org/obo/go/releases/2019-07-31/go.owl`
- Loaded from: `file:/projects/stars/jenkins-c8/workspace/GO/update-chebi-import/go-ontology/src/ontology/reasoned.owl`

#### Added
- [(-)-beta-caryophyllene](http://purl.obolibrary.org/obo/CHEBI_10357) SubClassOf [substance with insect attractant role](http://purl.obolibrary.org/obo/GOCHE_24850) 

#### Added
- [(6R)-5,10-methylenetetrahydrofolate(2-)](http://purl.obolibrary.org/obo/CHEBI_15636) SubClassOf [substance with cofactor role](http://purl.obolibrary.org/obo/GOCHE_23357) 

#### Added
- [(6R)-5,10-methylenetetrahydrofolic acid](http://purl.obolibrary.org/obo/CHEBI_1989) SubClassOf [substance with cofactor role](http://purl.obolibrary.org/obo/GOCHE_23357) 

#### Added
- [(6S)-5,6,7,8-tetrahydrofolate(2-)](http://purl.obolibrary.org/obo/CHEBI_57453) SubClassOf [substance with cofactor role](http://purl.obolibrary.org/obo/GOCHE_23357) 

#### Added
- [(6S)-5,6,7,8-tetrahydrofolic acid](http://purl.obolibrary.org/obo/CHEBI_15635) SubClassOf [substance with cofactor role](http://purl.obolibrary.org/obo/GOCHE_23357) 

#### Added
- [(R)-lipoate](http://purl.obolibrary.org/obo/CHEBI_83088) SubClassOf [substance with cofactor role](http://purl.obolibrary.org/obo/GOCHE_23357) 

#### Added
[...]

PR #19260, opened 2020-04-11
added chebi terms used in reactome missing from go-plus

goodb:
In order for the reacto ontology to work by importing the rest of go-lego (mainly go-plus), it needs these terms from chebi added into the chebi import.  If I remember correctly, this file is how that happens.


balhoff:
@goodb this is the correct approach, but unfortunately updating the CHEBI import has been blocked by needed revision to handling CHEBI roles (#17673).

goodb:
@balhoff okay, let me know if you need anything else on my end for this.  I guess I will assume that when that issue is resolved, this one will follow?

balhoff:
Yep.

PR #19288, opened 2020-04-18
add build for reacto.owl and go-lego-reacto.owl to stage_release

goodb:
addresses https://github.com/geneontology/pipeline/issues/35

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -71,6 +71,7 @@ GO_MAIN_RELATIONS = $(GO_BASIC_RELATIONS) BFO:0000051 BFO:0000066 RO:0002091 RO:
 GO_PLUS = extensions/go-plus
 GO_GAF = extensions/go-gaf
 GO_LEGO = extensions/go-lego
+GO_LEGO_REACTO = extensions/go-lego-reacto
 
 SPARQLDIR = ../sparql
 REPORTDIR = $(RELEASEDIR)/reports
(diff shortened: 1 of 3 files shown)


src/ontology/Makefile
```diff
@@ -216,6 +217,18 @@ $(GO_GAF).owl: extensions/go-gaf-edit.ofn $(GO_PLUS).owl extensions/gorel.owl mi
 $(GO_LEGO).owl: extensions/go-lego-edit.ofn $(GO_PLUS).owl mirror/ro-download.owl extensions/legorel.owl extensions/go-bfo-bridge.owl mirror/taxslim.owl mirror/taxslim-disjoint-over-in-taxon.owl
 	$(ROBOT) merge -i $< --collapse-import-closure true --include-annotations false reason -r ELK --exclude-duplicate-axioms true --remove-redundant-subclass-axioms true --exclude-tautologies structural relax reduce --named-classes-only true annotate -O $(BASE)/$@ -V $(RELEASE_URIBASE)/$@ -o $@
 
+extensions/reacto.owl: $(GO_LEGO).owl mirror/chebi.owl
+	mkdir -p reacto-tmp
+	wget -O ./reacto-tmp/biopax2go.jar --no-check-certificate https://github.com/geneontology/pathways2GO/releases/download/v0.1.0/biopax2go.jar
+	wget -O ./reacto-tmp/biopax.zip --no-check-certificate https://reactome.org/download/current/biopax.zip
+	unzip -o ./reacto-tmp/biopax.zip -d ./reacto-tmp/
+	java -jar -Xmx8G ./reacto-tmp/biopax2go.jar -b ./reacto-tmp/Homo_sapiens.owl -reacto ./extensions/reacto -lego $(GO_LEGO).owl -chebi ./mirror/chebi.owl
```
balhoff, line 225: Does `-jar` need to go just before the specified jar?

balhoff (changes requested):
I approve, just have one question inline.

src/ontology/Makefile
goodb: It works as is, tested.  I usually do it that way.  I guess it might also work with the memory allocation argument right after java ?

balhoff: Interesting, I just always thought that the jar file was an argument to `-jar`.

goodb: I think it is indeed an argument to -jar.  I guess its clever enough to deal with both.  I poked around a little and see examples of both patterns all over.

balhoff (approved):

kltm:
I think this will be fine. Let's try this out. Worst case, we separate out the target to something separate to handle on the side.

PR #19381, opened 2020-05-01
resolved differences between GO_0046932 and GO_0046962 per ticket 19379

hdrabkin:


hdrabkin:
David can you check my parentage change for these two

hdrabkin:
@ukemi I'm going to push this through; I can fix if you think I made an error here.

PR #19830, opened 2020-07-27
glycocalyx changes & new child terms, fixes #19782

### LINKED ISSUE #19782: issues with "glycocalyx" term, new terms needed?
I've just added a new term for `stereocilum coat` (aka `stereocilium glycocalyx`)  see #19780 

Thus, I've noticed that we have a term for `glycocalyx`, but it's part_of relationship makes it specific to bacteria, though it has been used for direct annotation of a mouse gene (part of sperm glycocalyx).

It has 2 child terms, 
- `capsule` where our definition specifies that it is present in fungi as well as in bacteria 
- `slime layer`where our definition only says bacteria, but Wikipedia says is this is in Archaea also

We also have a term for `lysosomal glycocalyx` (probably meant for eukaryotic lysosomes but has no annotations) which has no relationship to the existing `glycocalyx` term. 

It is clear that most metazoan cells have a glycocalyx (except it hasn't been observed in Trichoplax adhaerens and ctenophore). It seems likely that presence of glycocalyx is probably very ancient, though it is not present in Fungi or plant cells that have cell walls. 

Considering that the basic definition of glycocalyx isn't that different for prokaryotes and eukaryotes, and it is present in Archaea too, we could have a single term for `glycocalyx` and put terms like `capsule` and `stereocilium coat` directly under it. Or, if it's useful, we could have a top level term (which would then be not for annotation) and grouping terms for kingdoms: eubacteria, archaea, eukarota (or maybe just metazoa) which might be useful to provide more info about relationship to other cellular structures.

There are clearly many specific varieties of glycocaleces in metazoa, with at least some protein components that are specific to a given cell type and anatomical structure. Here are the ones I've come across:

- endothelial glycocalyx (GO:new?)
- lysosomal glycocalyx (GO:0090123, no annotations) => Kosicek 2018 PMID:29367433 Lamp1 & Lamp2
- platelet glycocalyx (GO:new?) - Nurden 2014 PMID:24967889 =>  GPIIb & GPIIIa
- sperm glycocalyx (GO:new?) - reannotate J:145640 => Defb126/Defb22
- stereocililum coat (GO:0120234) - Wu et al. 2019 PMID:31444330 => Pkhd1l1

krchristie:


krchristie:
This ticket also included addition of this term even though it doesn't show up in the diff in the `Files changed` tab. I had done a commit that I didn't push to origin and then done the rest of the work in a second commit and it seems that only the work in the second commit shows up in the diff.

`[Term]
id: GO:0120234
name: stereocilium coat
namespace: cellular_component
def: "A glycocalyx on the the endolymphatic surface of a cochlear hair cell that coats the external surface of each stereocilium and maintains a small distance between adjacent stereocilia in the bundle." [GOC:krc, ISBN:9781461268918, PMID:31444330, PMID:3583936]
synonym: "auditory hair cell glycocalyx" BROAD []
synonym: "stereocilium glycocalyx" EXACT []
is_a: GO:0030112 ! glycocalyx
relationship: part_of GO:0032420 ! stereocilium
created_by: krc
creation_date: 2020-07-15T00:56:35Z`

PR #19914, opened 2020-08-28
added relation fixes #19909

### LINKED ISSUE #19909: add regulates_o_transports_or_maintains_localization_of to GO-REL?
Although the chain relations in GO-REL will eventually be removed once annotations are imported into GO-CAM models. We have used a chain relation, regulates_o_transports_or_maintains_localization_of, that is not in GO-REL. Until the annotations are imported, they will fail new QC checks that determine if relations in extensions are in RO or GO-REL. Would it be possible to as this relation as a stop-gap to keep the annotations from being filtered?

@dougli1sqrd @kltm

pgaudet:


pgaudet:
Fixed the ID. It's too bad Protege doesn't support working on multiple ontologies - manually changing the idspace is tricky.

ukemi:
I agree working with more than one ontology in Protege is tricky. I've made this type of error in the past. This now looks ok to me. Should I go ahead and merge, or should we have @balhoff or @cmungall approve too?

ukemi (approved):

PR #19962, opened 2020-09-09
created new terms added subclasses fixes #19952

### LINKED ISSUE #19952: Group "complex terms" under nucleus
Would it be possible to group all of the complex terms directly under "nucleus" under a grouping term
"nuclear complex" (there is already a similar grouping term for "nuclear membrane complex"). I ask because there are so many complexes here that is is difficult to locate the nuclear parts hidden among them, and our users are not finding them......

This is  only a few:
Screenshot 2020-09-06 at 09 33 06

pgaudet:


pgaudet:
@valwood can you confirm that this is what you wanted ? It's a lot of changes to revert otherwise.

ValWood:
I prefer the grouping terms (if they are allowed). It will make scanning descendants much easier.

PR #19964, opened 2020-09-09
fixed intracellular term label, subclass fixes #17776

### LINKED ISSUE #17776: Intracellular children?
Currently the only subtype of this class is SMAD complex. Do we really mean for this term to represent an intracellular anatomical entitiy? We need to be able to define it more precisely.

pgaudet:


pgaudet:
@ukemi Can you check that this is right ?

PR #19987, opened 2020-09-15
resolve RHEA mapping to GO_0003955 by removing RHEA Creating new term…

hdrabkin:
… for NADP term and change parantage for NADH term


hdrabkin:
all wrong; killing this

PR #19355, opened 2020-04-28
Migrate all CHEBI references to pH 7.3 terms.

balhoff:
This is a version for review of GO with all references to CHEBI terms swapped with the pH 7.3 version as listed in Rhea's mapping file (see PR diff). Here is a report of lost and gained superclasses in the inferred GO hierarchy (all bio-chebi GCIs are removed from the version using pH 7.3 terms):

https://docs.google.com/spreadsheets/d/1saP7LyRWYLy6PGYlwO7UrHBaXH_8gwA5TxNpi9Fdelc/edit#gid=0

(first tab is relations lost, the second tab is relations added)

Issue #17763, opened 2019-08-26
L-beta-ethynylserine pathway


sylvainpoux:
Hi GO,
we would need a bunch of new terms for a new pathway

Thanks

Sylvain

Term:
L-beta-ethynylserine biosynthetic process

Definition:
The chemical reactions and pathways resulting in the formation of L-beta-ethynylserine, an antibiotic produced by Streptomyces bacteria.

Category: Process

child of:
GO:0017000 antibiotic metabolic process
GO:0008652 cellular amino acid biosynthetic process
GO:0043453 alkyne biosynthetic process

Xref
PMID:30867596
PMID:3082841
----------------------------------------------

Term:
L-propargylglycine biosynthetic process

Definition:
The chemical reactions and pathways resulting in the formation of L-propargylglycine (Pra), an antibiotic produced by Streptomyces bacteria.

Category: Process

child of:
GO:0017000 antibiotic metabolic process
GO:0008652 cellular amino acid biosynthetic process
GO:0043453 alkyne biosynthetic process

Xref
PMID:30867596
----------------------------------------------

Term:
L-propargylglycine synthase activity

Definition:
Catalysis of the reaction: L-2-amino-4-chloropent-4-enoate = chloride + H(+) + L-propargylglycine.

Category: Function

child of:
carbon-halide lyase activity ; GO:0016848

Xref
PMID:30867596
Rhea:59892
----------------------------------------------

Term:
L-propargylglycine--L-glutamate ligase activity

Definition:
Catalysis of the reaction: ATP + L-glutamate + L-propargylglycine = ADP + H(+) + L-gamma-glutamyl-L-propargylglycine + phosphate.

Category: Function

child of:
acid-amino acid ligase activity ; GO:0016881

PMID:30867596
Rhea:59896
----------------------------------------------

Term:
4-chloro-allylglycine synthase activity

Definition:
Catalysis of the reaction: 4-chloro-L-lysine + AH2 + O2 = A + formaldehyde + H2O + L-2-amino-4-chloropent-4-enoate + NH4(+).

Category: Function

child of:
oxidoreductase activity, acting on paired donors, with incorporation or reduction of molecular oxygen ; GO:0016705

Xref
PMID:30867596
Rhea:59888
----------------------------------------------

Term:
L-lysine 4-chlorinase activity

Definition:
[...]

PR #20093, opened 2020-10-08
Migrate CHEBI references to pH 7.3

balhoff:
Migrated all uses of CHEBI terms to the pH 7.3 version, wherever a mapping was provided by Rhea. Kept the biochebi GCIs in place for now. This supersedes #19355.

There is a small inference diff:

#### Removed
- [N(omega)-methyl-L-arginine metabolic process](http://purl.obolibrary.org/obo/GO_2001296) SubClassOf [cellular modified amino acid metabolic process](http://purl.obolibrary.org/obo/GO_0006575) 
- [N(omega)-methyl-L-arginine metabolic process](http://purl.obolibrary.org/obo/GO_2001296) SubClassOf [alpha-amino acid metabolic process](http://purl.obolibrary.org/obo/GO_1901605) 

#### Added
- [N(omega)-methyl-L-arginine metabolic process](http://purl.obolibrary.org/obo/GO_2001296) SubClassOf [organonitrogen compound metabolic process](http://purl.obolibrary.org/obo/GO_1901564) 

#### Removed
- [N(omega)-methyl-L-arginine catabolic process](http://purl.obolibrary.org/obo/GO_2001297) SubClassOf [cellular modified amino acid catabolic process](http://purl.obolibrary.org/obo/GO_0042219) 
- [N(omega)-methyl-L-arginine catabolic process](http://purl.obolibrary.org/obo/GO_2001297) SubClassOf [alpha-amino acid catabolic process](http://purl.obolibrary.org/obo/GO_1901606) 

#### Added
- [N(omega)-methyl-L-arginine catabolic process](http://purl.obolibrary.org/obo/GO_2001297) SubClassOf [organonitrogen compound catabolic process](http://purl.obolibrary.org/obo/GO_1901565)


krchristie:
Obviously this diff isn't very large, but I do have a concern about merging this in.

What would be the procedure going forwards for new terms that reference a ChEBI ID to get the proper reasoning of chemical parentage if we are using the 7.3 ChEBI terms in the equivalence axioms? For example, when I used the 7.3 ChEBI term `L-beta-ethynylserine zwitterion` instead of `L-beta-ethynylserine`, I ended up having to manually assert some parentage, which I think I should not have to do. In most cases, I'm not going to spend as much time as I did on this case to know what the parentage should be if it doesn't come in reasoning from the ChEBI term.

beta-ethynylserine-biosynProc


I don't really understand what the GCI's do, or how we will make sure that we get all the inferred reasoning over the chemical structure if we utilize the 7.3 terms directly in the equivalence axioms.

hdrabkin:
This looks like it’s because CHEBI is missing relationships. Would there be anything in place to catch new ones so that we can pass notice to CHEBI?
Do we even know the extent of this?
H

balhoff:
@krchristie I was hoping the GCIs would cover the differences between 7.3 and others. It seems like this is not always the case, given the 'N(omega)-methyl-L-arginine' stuff above, and also maybe your example.

What the GCIs do is say that having an input of X is logically the same as having an input of Y (where Y is is conjugate acid of X). So that helps it classify under things above either in the hierarchy.

I guess what I need to do is add 'has pH 7.3 form' to the list of properties that lead to GCI creation. But ideally we could drop the GCIs altogether and revise the CHEBI hierarchy (not feasible anytime soon I guess).

krchristie:
> This looks like it’s because CHEBI is missing relationships. Would there be anything in place to catch new ones so that we can pass notice to CHEBI? Do we even know the extent of this? H

According to a comment form @alanbridge (cc'd below) in the ticket  **L-beta-ethynylserine pathway** #17763  incomplete reasoning is quite common for the 7.3 forms. 

I don't think we should count on ChEBI to flesh out all the relationships for the 7.3 forms. Perhaps we should consider an option where we use the base compound in the axiom for reasoning and indicate the 7.3 form in some other way.

-K

> Hi Karen
> 
> the problem of inconsistent hierarchies in ChEBI between charge states is a common one.
> 
> But could you incorporate the ChEBI charge state mapping that Rhea provides into either
> 
> i) the GO or
> ii) the reasoning process
> 
> to deal with this?
> 
> This file - **chebi_pH7_3_mapping.tsv** - at https://www.rhea-db.org/download maps "other" charge states to that used in Rhea.
> 
> So for
> 
> CHEBI:144729 - L-β-ethynylserine zwitterion (used in Rhea)
> CHEBI:144833 - L-β-ethynylserine (the one with a better developed hierarchy)
> 
> the file gives something like this:
> 
> ```
> CHEBI	CHEBI_PH7_3	ORIGIN
> 144833	144729	computation
> ```
> 
> i.e. a computed relation between the two.
> 
> For reasoning you could infer relations using the hierarchy of both, but this could have some issues (like where one of the parents is also a specific charge form too). Perhaps another alternative could be to add **all** charge states from ChEBI as xrefs in the GO? So you would add ChEBI:144833 and ChEBI:144729 as xrefs for this pathway. Again for this one could use the chebi_pH7_3_mapping.tsv to help; if one GO xref has been curated for a GO BP, then additional GO xrefs could be added automatically from such a file.
> 
> Pinging @amorgat
> 
> All the best, Alan

ukemi:
Looks like at least in these to cases we can pin it down to the ChEBI difference in the tautomer versus conjugate acid/base relationship. For zwitterions, should we include the tautomer relation as well? @deustp01

balhoff:
Thanks @krchristie 

> I don't think we should count on ChEBI to flesh out all the relationships for the 7.3 forms. Perhaps we should consider an option where we use the base compound in the axiom for reasoning and indicate the 7.3 form in some other way.

I think my proposal to use the 7.3 relation to generate GCIs would provide this. Then we could keep the 7.3 forms in the regular GO axioms to have a better exact matchup with Rhea and Reactome.

ukemi:
PS. I don't think this change created this problem. I suspect it just illustrated it.

balhoff:
Should we discuss this next GO editors meeting? That will not be for a couple of weeks unfortunately. Or we can just meet anyway on Monday and discuss. :-)

krchristie:
> Thanks @krchristie
> 
> > I don't think we should count on ChEBI to flesh out all the relationships for the 7.3 forms. Perhaps we should consider an option where we use the base compound in the axiom for reasoning and indicate the 7.3 form in some other way.
> 
> I think my proposal to use the 7.3 relation to generate GCIs would provide this. Then we could keep the 7.3 forms in the regular GO axioms to have a better exact matchup with Rhea and Reactome.

That sounds good then. Basically, what I want as an ontology editor is to be able to use an appropriate ChEBI term in the equivalence axiom and NOT have to manually assert additional relationships that I would have gotten automatically if I had used the base compound, as demonstrated in the beta-ethynylserine example I included earlier. 

If we proceed with this migration to the 7.3 terms, will we be able to remove the manually asserted parentage for `L-beta-ethynylserine biosynthetic process` and have it be generated by the GCI? I don't know how to get to this version to test that myself.

krchristie:
> Should we discuss this next GO editors meeting? That will not be for a couple of weeks unfortunately. Or we can just meet anyway on Monday and discuss. :-)

I'm available Monday if you want to discuss this. @hdrabkin  said he could be available as well, even if he decides to take the day off.

krchristie:
> PS. I don't think this change created this problem. I suspect it just illustrated it.

Agreed, but we will create a different problem if we switch to the 7.3 forms without making sure that we can still get the chemical hierarchy reasoning that we have been getting from the base forms.

deustp01:
> I'm available Monday if you want to discuss this.

Me too. Maybe take the noon EDT / 9 AM PDT slot normally used for the weekly ontology call?

amorgat:
If you find it helpful to have someone from Rhea join the discussion, I'm available on Monday as well. Just tell us!

deustp01:
>  For zwitterions, should we include the tautomer relation as well?

As far as I can tell, yes. Formally, tautomers are isomers of a compound which exist together in equilibrium, and are readily interchanged by migration of an atom or group within the molecule, thus uncharged glycine is a tautomer of glycine zwitterion because moving a hydrogen atom from the COOH group to the NH2 group of glycine (CHEBI:15428) yields glycine zwitterion (CHEBI:57305).

(Obsessive weedy issue, not needed to decide what to do here. The intramolecular interchange bit is part of the dictionary definition but I'm not sure it's required - formally, it describes glycine zwitterion formation from glycine as intramolecular migration of a hydrogen but is that really how the the typical molecule of zwitterion forms? If one dissolved a mixture of glycine doubly uniformly labeled with C14 and H3, with unlabeled glycine in water, would all molecules of zwitterion be doubly labeled or unlabeled, or would some have only H3 because a proton dissociated from a labeled glycine associated with the NH2 group of a previously unlabeled glycine?)

And the outgoing links for for the two CHEBI entries highlight Karen's point: glycine has a full set including conjugate acid and base ones as well as the tautomer one, while the zwitter ion entry has only a link to glycine. This is particularly a problem because the predominant form of glycine molecule under physiological conditions is likely to be the zwitterion and it's hard to see how any substantial proportion of uncharged glycine could exist in aqueous solution under any conditions. So we annotate to a form that is poorly equipped to support reasoning.

That all makes the file Alan mentioned above - "This file - chebi_pH7_3_mapping.tsv - at https://www.rhea-db.org/download maps "other" charge states to that used in Rhea." look very attractive. What relative amounts of effort are needed to maintain GCIs versus this file versus ontology development in CHEBI itself? @alanbridge @balhoff

ukemi:
Always enjoy the weeds @deustp01. So I think we agree that we should have included the tautomer relation in the GCIs. The only downside to using a manually generated file for the charge state mappings is that it has to be maintained by hand. Our attempts using the conjugate-acid conjugate-base traversal was an attempt to make this more automated. Of course ideally if we had classes for all of the rungs of the charge state ladder up and down ChEBI, I believe the reasoning would be complete.

alanbridge:
Hi Peter and all, 

the file "chebi_pH7_3_mapping.tsv" - at https://www.rhea-db.org/download - maps "other" charge states to that used in Rhea. We create this file at each release, it's part of the Rhea production process and will continue to be available; users need it to map ChEBI to Rhea... 

For glycine, this file seems to map the form of glycine with no charge (CHEBI:15428), and the -ve and +ve charged forms, to the glycine zwitterion OK (CHEBI:57305).

15428	57305	computation
32507	57305	computation
32508	57305	computation
57305	57305	computation

Anne can give you more detail but it may well help address mapping of tautomers. 

In the GO you could use this file to either i) add additional xrefs to GO terms (e.g. make the GO charge agnostic and add them all?!) or ii) support reasoning by filling in links from "the major microspecies at pH 7.3 as used in Rhea"  to the "better annotated forms" in the sense of having a more developed set of parent classes.

All the best, Alan

amorgat:
ChEBI entries describing the different protonation states of a molecule are linked by **"is conjugate base of", "is conjugate acid of" and "is tautomer of" relationships**. 
Some relationships may be missing and retrieving all species maybe tricky when they have several ionizable groups (need of a recursive procedure). 
That's why instead of these 3 relationships, the SIB developers use an additional relationship "has_major_microspecies_at_pH_7_3" which is available in the chebi.owl file (published by Rhea) or the "chebi_pH7_3_mapping.tsv" file mentionned by @alanbridge .
The relationship is computed using the Marvin plugin provided by Chemaxon (https://docs.chemaxon.com/display/docs/Major+Microspecies+Plugin)

balhoff:
Plan: update biochebi tool to use pH 7.3 mapping for outputting GCIs, instead of the current acid/base relations. Then redo this to check diff.

balhoff:
I've reworked bio-chebi to base it on the pH7.3 mappings. Now I need some help; there is one unsatisfiable class which is both an `anion channel activity` and a `cation channel activity`:

## [ammonium channel activity](http://purl.obolibrary.org/obo/GO_0015251) EquivalentTo [Nothing](http://www.w3.org/2002/07/owl#Nothing) ##

  - [ammonium channel activity](http://purl.obolibrary.org/obo/GO_0015251) EquivalentTo [channel activity](http://purl.obolibrary.org/obo/GO_0015267) and ([transports or maintains localization of](http://purl.obolibrary.org/obo/RO_0002313) some [ammonium](http://purl.obolibrary.org/obo/CHEBI_28938))
    - [ammonium](http://purl.obolibrary.org/obo/CHEBI_28938) SubClassOf [monovalent inorganic cation](http://purl.obolibrary.org/obo/CHEBI_60242)
      - [monovalent inorganic cation](http://purl.obolibrary.org/obo/CHEBI_60242) SubClassOf [inorganic cation](http://purl.obolibrary.org/obo/CHEBI_36915)
  - [anion channel activity](http://purl.obolibrary.org/obo/GO_0005253) EquivalentTo [channel activity](http://purl.obolibrary.org/obo/GO_0015267) and ([transports or maintains localization of](http://purl.obolibrary.org/obo/RO_0002313) some [inorganic anion](http://purl.obolibrary.org/obo/CHEBI_24834))
  - [anion channel activity](http://purl.obolibrary.org/obo/GO_0005253) DisjointWith [cation channel activity](http://purl.obolibrary.org/obo/GO_0005261)
  - [monovalent inorganic anion](http://purl.obolibrary.org/obo/CHEBI_79389) SubClassOf [inorganic anion](http://purl.obolibrary.org/obo/CHEBI_24834)
  - [cation channel activity](http://purl.obolibrary.org/obo/GO_0005261) EquivalentTo [channel activity](http://purl.obolibrary.org/obo/GO_0015267) and ([transports or maintains localization of](http://purl.obolibrary.org/obo/RO_0002313) some [inorganic cation](http://purl.obolibrary.org/obo/CHEBI_36915))
  - [azanide](http://purl.obolibrary.org/obo/CHEBI_29337) SubClassOf [monovalent inorganic anion](http://purl.obolibrary.org/obo/CHEBI_79389)
  - [transports or maintains localization of](http://purl.obolibrary.org/obo/RO_0002313) some [ammonium](http://purl.obolibrary.org/obo/CHEBI_28938) EquivalentTo [transports or maintains localization of](http://purl.obolibrary.org/obo/RO_0002313) some [azanide](http://purl.obolibrary.org/obo/CHEBI_29337)

What needs to change here?

pgaudet:
azanide and ammonium should not be considered equivalent - is this what is happening here ? 

Ammonium is a cation (+ charge), azanide an anion (- charge). 

CHEBI has 'azanide (CHEBI:29337) is conjugate base of ammonia (CHEBI:16134) ', which is probably correct ?

balhoff:
Thanks, it looks like the old bio-chebi makes use of a `no_conj_equiv` tag on 4 CHEBI terms (including ammonium and azanide) to exclude certain GCIs. I can replicate this, but what is the basis for excluding only these terms?

ukemi:
I believe it is because some of the conjugate_acid/base species in ChEBI are never present physiologically. Ammonia and its conjugate acids and base have always been a problem. I don't think azanide would ever exist under physiological conditions. See sodium amide.

ukemi:
I believe this was done after the work in the paper when subsequent unsatisfiable classes were found due to the additions to ChEBI. I remember discussing it, so there is probably something in an old ticket or in the wiki. Note that we also say that we didn't calculate GCIs for tautomers in the paper. I also think that the issues we discussed on the call didn't come up until later, when ChEBI was made more complete.

hdrabkin:
See also
https://www.wikigenes.org/e/chem/e/2826723.html

deustp01:
> See also
> https://www.wikigenes.org/e/chem/e/2826723.html

But this looks like a mistake - the article appears to be equating an amino group (R-NH2), e.g., the amino terminus of a protein, with the free species NH2-.

amorgat:
When I load CHEBI:29337 in MarvinSketch (ChemAxon) and run the pKa plugin, the application only shows ammonia (CHEBI:16134) and ammonium (CHEBI:28938). No issue related to the procedure itself.

I sent a mail to Adnan and Venkat and advertise this ticket.

Capture d’écran 2020-10-20 à 11 40 04

ukemi:
One thing that might be very useful is to tag all species that would never be found under physiological conditions. Somehow I remember a similar discussion about elemental sodium.

hdrabkin:
I remember the discussion about elemental metals.

deustp01:
> tag all species that would never be found under physiological conditions

But carefully. No one is trying to annotate extremophile biochemistry for GO, but we probably don't want to close that door completely. But I guess this about the same as saying the pH 7.3 feature and its extensions may someday need a taxon constraint. That might be OK.

alanbridge:
Organelle pH can vary pretty dramatically too, but in Rhea we decided to stick with one charge state to simplify life.

ukemi:
That was our decision as well.

balhoff:
To make sure it's clear what I'm doing, I've replaced the old bio-chebi script with one based on the Rhea pH 7.3 mappings. It seems like it's almost completely the same result (but I didn't yet do the checks to see if it fills the gaps created by swapping in the 7.3 terms everywhere), but initially I didn't implement the small `no_conj_equiv` exclusion list. For that list, it prevents use of mappings involving those terms to generate GCIs. You can see an example GCI created from the pH 7.3 mappings, in the OWL explanation above:

```
transports or maintains localization of some ammonium EquivalentTo transports or maintains localization of some azanide
```

If I add the exclusion list, this GCI will no longer be created. So that will solve this logical problem. But I'm kind of curious if we should just continue using this small exclusion list or if there is some kind of ontology-based/data-driven approach to use.

cmungall:
The `no_conj_equiv` approach feels unsatisfactory. It is a semi-secret exclusion rule on an already baroque series of workarounds. 

How many conjugate relation chains span anion to cation? Does this correspond to the manual exclusion list?

balhoff:
Here is the full exclusion list:

- [azanide](http://purl.obolibrary.org/obo/CHEBI_29337)
- [ammonia](http://purl.obolibrary.org/obo/CHEBI_16134)
- [ammonium](http://purl.obolibrary.org/obo/CHEBI_28938)
- [hydridonitrate(2−)](http://purl.obolibrary.org/obo/CHEBI_29340)

balhoff:
The classification of 'N(omega)-methyl-L-arginine metabolic process' and 'N(omega)-methyl-L-arginine catabolic process' is restored now with the bio-chebi revision.

balhoff:
> How many conjugate relation chains span anion to cation? Does this correspond to the manual exclusion list?

@cmungall I think the answer is around 17,000. Is this what you mean? `cation and 'is conjugate acid of' some anion`

PR #20206, opened 2020-10-26
update reacto build id

goodb:
Updating URL where jar file needs to be collected from. 
https://github.com/geneontology/pathways2GO/releases/tag/v1.0.0 

(Todo would be nice to build it from the source. At some point we should figure out why that wasn't working on the server..)


kltm:
I'll assume this works; we can revert if there is a problem.

PR #20212, opened 2020-10-27
Update x_maturation.yaml

pgaudet:
I see 'results in maturation of: "RO:000229" used in the ontology, not "results in developmental progression"
Also, the tsv file is empty, suggesting that this design pattern is not correct.


balhoff (changes requested):
@pgaudet you will also need to change the property used in the equivalent class expression down below.

pgaudet:
oh yes thanks !

PR #20321, opened 2020-11-10
Create mf_binding.yaml

pgaudet:


src/design_patterns/mf_binding.yaml
```diff
@@ -0,0 +1,50 @@
+pattern_name: 'activator_activity'
```
cmungall, line 1: oops!

src/design_patterns/mf_binding.yaml
```diff
@@ -0,0 +1,50 @@
...
+  molecular_function: 'GO:0003674'
+  binding: 'GO:0005488'
+
+relations:
+  has input: 'RO:0002629' 
+  
+vars:
+  has_input: %s 
+
+data_list_vars:
```
cmungall, line 13: @balhoff do we need all this boilerplate? can we not simplify these like we have e.g. for mondo?

balhoff: No, we don't need the boilerplate. That is a holdover from the old patterns David had made.

pgaudet:
Is this better ? I removed both 'boiler plate' sections. I can remove from all design pattern files, if that's OK.

src/design_patterns/mf_binding.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+classes:
+  molecular_function: 'GO:0003674'
+  binding: 'GO:0005488'
+
+relations:
+  has input: 'RO:0002629' 
+  
+vars:
+  has_input: %s
```
balhoff, line 11: Instead of `%s` here, you need to put the type, e.g. `"'molecular_function'"`

src/design_patterns/mf_binding.yaml
```diff
@@ -0,0 +1,32 @@
...
+  has input: 'RO:0002629' 
+  
+vars:
+  has_input: %s 
+
+data_list_vars:
+  def_dbxrefs: "xsd:string"
+  
+name:
+  text: "%s binding"
```
balhoff, line 17: Need a `vars:` list.

src/design_patterns/mf_binding.yaml
```diff
@@ -0,0 +1,32 @@
...
+    
+def: 
+  text: "Binding to a %s."
```
balhoff, line 20: Need a `vars:` list.

src/design_patterns/mf_binding.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# Boilerplate - don't edit
+xref: 
+   value: cross_references
+
+namespace: 
+  text: 'biological_process'  # e.g. biological_process, quoting is optional
+
+equivalentTo: 
+  text: "'binding' has input some %s"
```
balhoff, line 31: Need a `vars:` list.

balhoff, line 31: Some syntax fixes for expression:

`"'binding' and ('has input' some %s)"`

balhoff (changes requested):

PR #20633, opened 2020-12-23
fetch mappings of KEGG.ORTHOLOGy to GO

cmungall:


balhoff (changes requested):
Seems to work but I don't think the formatting is quite right:

```
"x"	"y"
KEGG.ORTHOLOGY:K03919	GO:0103053
```

vs.

```
! Generated on 2018-06-26T16:02Z from the ontology 'go' with data version: 'releases/2018-05-14'
!
EC:1 > GO:N-ethylmaleimide reductase activity ; GO:0008748
EC:1 > GO:oxidoreductase activity ; GO:0016491
```

PR #20635, opened 2020-12-27
Remove definition dbxref that is also an alt_id for GO:0002949

kimrutherford:
This seems like a mistake?  This term has its own alt_id as a dbxref for the definition.


pgaudet:
@kimrutherford Not sure why your PR didn't work. I fixed this here: https://github.com/geneontology/go-ontology/pull/20645

Thanks, Pascale

balhoff:
@pgaudet the PR did work — it's just that now there are 5 checks instead of 1, to support the inference diff. Only the QC check is required to pass for merge. Right now there is a problem with the inference diff not being able to comment on PRs opened from outside repos (in this case @kimrutherford has a fork instead of a branch in this repo).

I will be trying to find a work around for that issue, but it will come up only rarely since we don't get many PRs from forks.

pgaudet:
Thanks for the explanation!

PR #20726, opened 2021-01-21
Create transcription_factor_activity.tsv

pgaudet:


pgaudet:
Hi @balhoff 
Please have a look. 

I was not able to make a textual definition in which we can just substitute other classes. The parent class (GO:0003700) is defined as "A transcription regulator activity that modulates transcription of gene sets via selective and non-covalent binding to a specific double-stranded genomic DNA sequence (sometimes referred to as a motif) within a cis-regulatory region. Regulatory regions include promoters (proximal and distal) and enhancers. Genes are transcriptional units, and include bacterial operons."

and children are simpler, for example for activator (GO:0001216) "A DNA-binding transcription factor activity that activates or increases transcription of specific gene sets."

Thanks, Pascale

balhoff (changes requested):
In the logical definition, you are using two variables. These need to come from the names defined under the top-level `vars` key:

```yaml
vars:
   biological_process: "'regulation of transcription, DNA-templated'"
```

The names you're referencing are the term "vocabulary" defined for this file (`classes`). So instead try defining `var`s naming the two roles these variables are playing in the logical definition.

PR #21150, opened 2021-03-19
Hechtian strand new term, fixes #21138

### LINKED ISSUE #21138: hechtian strand, new term request
Dear GO-Team,
there is a localization for enzyme 2.4.1.173: "hechtian strand". I found this definition in Wikipedia: A stretched plasma membrane extending from the plasmolysed protoplast to the cell wall in plants. (https://en.wiktionary.org/wiki/Hechtian_strand).
Could it be a new term, or would you use the term "plasma membrane" instead?
Thanks for your help.
Kind regards, Marion Gremse (from brenda-enzymes.org)

krchristie:


balhoff:
@krchristie there is a problem with the taxon constraint (see #21151). 'contractile fiber' is only in Metazoa.

krchristie:
> @krchristie there is a problem with the taxon constraint (see #21151). 'contractile fiber' is only in Metazoa.

@balhoff - The real error here is quite simple. For some reason, Protege doesn't let me create new terms, so I usually duplicate a term with somewhat appropriate parentage and edit it. In this case, I forgot to remove a couple part_of relationships that belonged to the term I duplicated, but shouldn't have been kept on this new term.

I've deleted these relationships since they shouldn't have been on this term anyway:
- 'part of' some 'contractile fiber'
- 'part of' some sarcomere

I suppose at some point I should see if you can help me figure out why the new term buttons don't work for me now.

pgaudet:
@krchristie you must be in the 'asserted' view, and not 'inferred' to create a new term. 
Is this your problem?

krchristie:
> @krchristie you must be in the 'asserted' view, and not 'inferred' to create a new term.
> Is this your problem?

Yes, that is my problem. Has it always been that way? I've used the Inferred view as my preferred view for a long time. It also seems like a design flaw to keep those icons present in a view where they do not work and do not generate an error message either. I wonder why they don't work in the Inferred view. It works fine to duplicate a term in that view. More generally, I wonder what the rationale to try to restrict term creation to the Asserted view is. I duplicate terms in the Inferred View all the time and it seems fine...

pgaudet:
It's been like that at least since I joined the group in 2017 - see the Guidelines we wrote following the Berkeley editors meeting 
http://wiki.geneontology.org/index.php/Guidelines_for_creating_a_GO_term
I thought you were there too ? 

In any case, if there are other problematic issues please bring them up at the ontology call. 

Thanks, Pascale

krchristie:
> It's been like that at least since I joined the group in 2017 - see the Guidelines we wrote following the Berkeley editors meeting
> http://wiki.geneontology.org/index.php/Guidelines_for_creating_a_GO_term
> I thought you were there too ?
> 
> In any case, if there are other problematic issues please bring them up at the ontology call.
> 
> Thanks, Pascale

I was there, but don't remember this issue. I guess I was making my terms in the Asserted view for a while. When I switched to using the Inferred View all the time, it was only mildly annoying since I got into the habit of making my terms by duplicating an existing term. Sometimes that's actually easier, especially when I'm putting in several terms that are related using the same papers as definition references.

PR #21531, opened 2021-05-29
disambiguate optic placode terms Fixes #12652

### LINKED ISSUE #12652: optic lobe placode formation
The definition of this term looks like a confused mix of vertebrate optic placode and insect optic _lobe_ placode.

"optic lobe placode formation"
definition "Establishment of the optic placode, a thickened area of densely packed ectoderm cells directly overlying the optic vesicle in the early embryo. In Drosophila, for example, the placode appears in the dorsolateral region of the head in late stage 11 embryos and is the precursor to the larval visual system."^^xsd:string

First sentence  is vertebrate (it even says optic placode) - optic vesicle is not an insect struc.  

Should probably be split, but this will require some related work in Uberon see obophenotype/uberon#135

raymond91125:


pgaudet (commented):
Hi @raymond91125 
I dont see in the ticket that we wanted to use the word 'insect' in the term labels. Actually we do try to avoid that. Maybe I missed a discussion?

raymond91125:
"insect visual primordium" is the label for the Uberon term in logical def
 thus I used it.

pgaudet:
OK, sounds good, please go ahead and merge.

Thanks, Pascale

Le mer. 2 juin 2021 à 7:47 PM, Raymond Lee ***@***.***> a
écrit :

PR #21566, opened 2021-06-07
clarify definitions of "developmental process involved in reproductio…

### LINKED ISSUE #12332: 'developmental process involved in reproduction'  & 'reproductive structure development'
These two terms sound very similar. Should they be merged?

label "reproductive structure development"^^string
definition "The reproductive developmental process whose specific outcome is the progression of structures that will be used in the process of creating new individuals from one or more parents, from their formation to the mature structures."^^string
EquivalentTo: 'anatomical structure development' and ('results in development of' some 'reproductive structure')

label "developmental process involved in reproduction"^^string
definition "A developmental process in which a progressive change in the state of some part of an organism specifically contributes to its ability to form offspring."^^string
EquivalentTo: 'developmental process' and ('part of' some reproduction)

For the second one, the EC def doesn't match the textual def. This needs to be fixed whether or not the terms are merged.

raymond91125:
…n" and "reproductive structure development", Fixes #12332


raymond91125:
@pgaudet Is this narrow solution good to go?

pgaudet:
Yes, this looks good to me, thanks.

PR #21593, opened 2021-06-11
NTR purine ribonucleotide salvage  GO_0106380 per ticket 21581

hdrabkin:


hdrabkin:
@ukemi can you just check to see if I did the log def correctly (and the asserted parent) before I proceed?

PR #21664, opened 2021-06-20
TCs for echolocation

cmungall:
TCs for echolocation

 - only in mammals
 - adding never-ins to ensure these don't appear in M musculus and primate subsets
 - adding present ins for bats and toothed whales (well known) and soft-furred tree mices (new: https://doi.org/10.1126/science.aay1513)

question: how do we record provenence in TCs now we use TSVs?﻿


cmungall:
Replaced by #21679

Issue #21572, opened 2021-06-08
GO:0140030 	modification-dependent protein binding


hattrill:
I think that it would be good to review the use of these terms:

GO:0072572	poly-ADP-D-ribose binding
GO:0072571	mono-ADP-D-ribose binding
GO:0072570	ADP-D-ribose binding

![Screenshot 2021-06-08 at 12 26 19](https://user-images.githubusercontent.com/11865746/121177896-d4a09080-c855-11eb-9df8-7cdf9778df22.png)

I think that the majority would apply to the binding of proteins PTM'd by ADP-D-ribose(s) rather than the free unit. PMID:20088964

An example are Ub ligases that target ADP-ribosylated proteins. e.g. http://www.ebi.ac.uk/interpro/entry/InterPro/IPR033509/
http://www.pantree.org/node/annotationNode.jsp?id=PTN001002511

hattrill:
directly relates to #12787

ValWood:
Should we even have terms for 'modification -dependent protein binding'?
We have 2 ways to capture this. One with a term, and one with a modified form. Would it be better if we always did it one way?

hattrill:
As these are binding a motif + mod, it's a bit more expansive than a one:one relationship that is captured  by mod protein binding + PRO id 1:1.
I think having higher order classes, such as 'glycosylation-dependent protein binding' is also great when there is a good chance that you don't know the exact nature of the mod or even where it is.

ValWood:
> As these are binding a motif + mod, it's a bit more expansive than a one:one relationship that is captured by mod protein binding + PRO id 1:1.

I didn't get this part

>I think having higher order classes, such as 'glycosylation-dependent protein binding' is also great when there is a good chance that you don't know the exact nature of the mod or even where it is.

Yes that's a good point.  But I think (although am not 100% sure) that you can have a generic modified form from PRO.

hattrill:
What I mean is that the ADP-ribs. motif may reside in any number of different proteins (https://en.wikipedia.org/wiki/ADP-ribosylation). Querying based on this term would have use cases.

We've been here and dicussed this before: #12787 and these grouping term were deemed useful. Looks to me as if the ADP-D-ribose binding terms were missed in this review.

raymond91125:
Existing annotations (experimental evidence)
Term | GP | Evidence (PMID) | binding to (ribose / ribosylated protein)
[...]

PR #21838, opened 2021-07-12
Added three ADP-D-ribose modification-dependent protein binding terms…

raymond91125:
…. #21572


raymond91125:
@pgaudet good to merge?

PR #21864, opened 2021-07-21
added nucleolar chromatin organization partof nucleolus organization.…

### LINKED ISSUE #10746: query: nucleolus organization
Is this really a "process" or is it only a phenotype that is a consequence of problems with
rRNA processing/ ribosome biogenesis?

The gene products annotated to it (only ~24 experimental appear to be defects in these processes).

If it is a 'good' term, should it have things like:
protein localization to nucleolar rDNA repeats 
recombination within rDNA repeats 
replication fork processing at rDNA locus 
rDNA condensation 
be children?

Reported by: ValWood

Original Ticket: [geneontology/ontology-requests/10559](https://sourceforge.net/p/geneontology/ontology-requests/10559)

raymond91125:
… Fixes #10746


ValWood:
Yes, thanks!

Issue #22374, opened 2021-11-10
NTR: [mycophenolic acid biosynthetic process]


marcfeuermann:
I would propose the creation of a BP term linked to mycophenolic acid biosynthesis.

"mycophenolic acid biosynthetic process"

I would only create the biosynthesis terms since in general there are no degradation pathways of these metabolites (only few exceptions).

Definition:
The chemical reactions and pathways resulting in the formation of mycophenolic acid (MPA), the first isolated antibiotic natural product in the world obtained from a culture of Penicillium brevicompactum in 1893.

PMID: 26751579, 31209052, 21398490

Synonyms: mycophenolic acid synthesis, MPA synthesis, mycophenolic acid biosynthesis, MPA biosynthesis, mycophenolic acid formation, MPA formation, mycophenolic acid anabolism, MPA anabolism.

Proteins to be annotated to this term:

PMID | Protein | UniProt | Type | Rhea

26751579,   31209052, 21398490 | mpaC, mpaC' | W6QL47, F1DBA9, A0A0B5KU17 | Cytosolic polyketide synthase   mpaC' | RHEA:63056

26751579, 31209052, 22544261 | mpaDE, mpaDE' | W6QRN8, I7LRH3, A0A0B5KYT4 | ER-bound cytochrome P450 monooxygenase | RHEA:66669, RHEA:66673

31209052, 25630520 | mpaA, mpaA' | W6QIT8, F1DBA7, A0A0B5L778 | Golgi apparatus-associated prenyltransferase | RHEA:66677
31209052 | mpaB, mpaB' | W6R4H8, F1DBA8, A0A0B5LB52 | ER-bound oxygenase | RHEA:66688
31209052, 25630520 | mpaG, mpaG' | W6R4H4, F1DBB3, A0A0B5L781 | O-methyltransferase | RHEA:66696
31209052 | mpaH, mpaH' | W6QL41, F1DBB4, A0A0B5LB55 | Peroxisomal acyl-CoA hydrolase | RHEA:66704
31209052 | AC891 | W6R1D9, P9WEY3 | Acyl-CoA ligase | RHEA:66700
26751579 | mpaF, mpaF' | W6QIT2 , F1DBB2, A0A0B5L585 | Inosine-5'-monophosphate dehydrogenase | RHEA:11708

Parent terms: 
GO:0030639 polyketide biosynthetic process
GO:0017000 antibiotic biosynthetic process
GO:0043386 mycotoxin biosynthetic process

[...]

Issue #22373, opened 2021-11-10
NTR: [patulin biosynthetic process]


marcfeuermann:
I would propose the creation of several BP terms linked to secondary metabolites biosynthesis.

The first one would be:

"patulin biosynthetic process"

I would only create the biosynthesis terms since in general there are no degradation pathways of these metabolites (only few exceptions).

Definition:
The chemical reactions and pathways resulting in the formation of patulin, an acetate-derived tetraketide mycotoxin produced by several fungal species that shows antimicrobial properties against several bacteria.

PMID: 30680886, 25120234

Synonyms: Patulin synthesis, patulin biosynthesis, patulin formation, patulin anabolism

Proteins to be annotated:

PMID | Protein | UniProt | Type | Rhea

30680886 | patK | A0A075TRC0 | NR-PKS | RHEA12240
30680886, 24334092 | patG | A0A075TXZ1, A1CFL4 | 6-methylsalicylic acid decarboxylase | RHEA:23112
30680886 | patH | A0A075TRL5 | Cytochrome P450 monooxygenase | RHEA:62208
30680886 | patI | A0A075TMP8 | Cytochrome P450 monooxygenase | RHEA:62212
30680886 | patJ | A0A075TR41 | Oxydoreductase |  
30680886 | patO | A0A075TR33 | FAD-linked oxidoreductase |  
30680886 | patN | A0A075TRB3 | Isoepoxydon dehydrogenase | RHEA:62216
30680886 | patF | A0A075TR27 | ? | RHEA:62220
30680886 | paD | A0A075TMP0 | Alcohol dehydrogenase | RHEA:62224
30680886 | patE | A0A075TRK9 | Dehydrogenase | RHEA:62228
30680886 | patB | A0A075TXZ3 | Carboxylesterase | Role unknown
25120234 | patM | B6RAL1 | ABC transporter |  
25120234 | patA | A0A075TRL0 | Acetate transporter protein |  

Parent terms: 
GO:0030639 polyketide biosynthetic process
GO:0043386 mycotoxin biosynthetic process

Child terms: (to create)
Regulation of patulin biosynthetic process
Negative regulation of patulin biosynthetic process
Positive regulation of patulin biosynthetic process

Protein that can be annotated with positive regulation of patulin biosynthetic process:
PMID:25625822 | patL | A0A075TXZ8 | Transcription regulator

Taxon constraint: Only in taxon 4751 Fungi

[...]

PR #22404, opened 2021-11-18
added terms for #22374 for #22373

pgaudet:


pgaudet:
@balhoff We can merge the PR with these errors?

balhoff:
Tricky error to find in the log:

```
ERROR: CHEBI IDs should be mapped to pH 7.3:
168396
```

CHEBI:168396 is not in the pH 7.3 set.

pgaudet:
Thanks @balhoff 
A side question: CHEBI: 168396 is not in the imports, ie it's somehow imported automatically. Could we make it that only allowed CHEBI chemicals would be imported? or available to GO editors? 

Thanks, Pascale

Issue #22414, opened 2021-11-18
NTR: GC-box binding


RLovering:
Please provide as much information as you can: 

* **Suggested term label:** GC-box binding

* **Definition (free text)**  Binding to an GC-box, a DNA motif with the consensus sequence GGGCGG that is located upstream of the start point of eukaryotic transcription units. The GC-box may occur in multiple copies or in either orientation.

* **Reference, in format PMID:#######** PMID: 14701757 and http://purl.obolibrary.org/obo/SO_0000173

* **Gene product name and ID to be annotated to this term** human SP1 P08047

* **Parent term(s)**  GO:0000978 RNA polymerase II cis-regulatory region sequence-specific DNA binding

* **Children terms (if applicable)** Should any existing terms that should be moved underneath this new proposed term? None

* **Synonyms (please specify, EXACT, BROAD, NARROW or RELATED)**
* EXACT: GC box binding
* BROAD: GC rich promoter region binding
* RELATED: GC-rich region binding
* RELATED: GC_rich_promoter_region binding
* Sorry I can never remember which type of synonym these would be - please correct as required

* **Cross-references**
* For enzymes, please provide RHEA and/or EC numbers. 
* Can also provide MetaCyc, KEGG, Wikipedia, and other links.

* **Any other information**
Note that the parent term is not promoter binding, even though this is the parent in SO, as we do not have promoter binding in GO only core-promoter binding which is too specific for the GC-box binding term

pgaudet:
+[Term]
+id: GO:0140728
+name: GC-box binding
+def: "Binding to a GC-box, a DNA motif with the consensus sequence GGGCGG that is located upstream of the start point of eukaryotic transcription units. The GC-box may occur in multiple copies or in either orientation relative to the transcription start site." [PMID:14701757]
+synonym: "GC box binding" EXACT []
+synonym: "GC rich promoter region binding" BROAD []
+synonym: "GC-rich region binding" BROAD []
+synonym: "GC_rich_promoter_region binding" BROAD []
+is_a: GO:0000978 ! RNA polymerase II cis-regulatory region sequence-specific DNA binding
+property_value: term_tracker_item https://github.com/geneontology/go-ontology/issues/22414 xsd:anyURI
+created_by: pg
+creation_date: 2021-11-23T08:51:58Z
+

pgaudet:
Need to add LD when SO imports have run

PR #22434, opened 2021-11-23
For #22414

pgaudet:


pgaudet:
@balhoff can you please regenerate SO imports?

PR #22509, opened 2021-12-02
Enable property path reasoning for never_in_taxon constraints.

balhoff:
See issue #3942.

PR #22511, opened 2021-12-03
Removed GO:0019933	& Dictyostelids

pgaudet:
cAMP-mediated signaling SubClassOf in taxon some (not (Dictyostelia)) is WRONG 
"adenylate cyclase-activating G protein-coupled cAMP receptor signaling pathway " uses cAMP as both intra and extracellular signaling


pgaudet:
For https://github.com/geneontology/go-ontology/pull/22509

PR #22517, opened 2021-12-06
Switch to uberon-base for import.

balhoff:


balhoff:
Lots of uberon inferences in the inference diff is okay for now. Should go away if we move to a single mega-import.

balhoff:
Putting on hold; will work on mega-import due to missing inferences.

Issue #19947, opened 2020-09-06
NTR mitotic nuclear pore complex disassembly


ValWood:
NTR mitotic nuclear pore complex disassembly

definition
The mitotic cell cycle process in which the controlled breakdown of the nuclear pores occurs during open or closed mitosis. 

Descendant of
"mitotic nuclear envelope disassembly"

synonym  "nuclear pore complex disassembly during mitosis"

def
Closed mitosis requires local disassembly of the nuclear envelope.
Dey G, Culley S, Curran S, Schmidt U, Henriques R, Kukulski W, Baum B.
Nature. 2020 Sep;585(7823):119-123. doi: 10.1038/s41586-020-2648-3. Epub 2020 Aug 26.
PMID: 32848252

pgaudet:
id: GO:0140516
name: mitotic nuclear pore complex disassembly
namespace: biological_process
def: "The mitotic cell cycle process in which the controlled breakdown of the nuclear pores occurs during open or closed mitosis." [PMID:32848252]
synonym: "nuclear pore complex disassembly during mitosis" EXACT []
is_a: GO:0007077 ! mitotic nuclear envelope disassembly
property_value: term_tracker_item https://github.com/geneontology/go-ontology/issues/19947 xsd:anyURI
created_by: pg
creation_date: 2020-09-09T15:29:01Z

PR #22531, opened 2021-12-07
added disjoint statement

pgaudet:


cmungall:
Great start!

in this case, this seems to be a typo rather than something asserted by design:

id: GO:0140516
name: mitotic nuclear pore complex disassembly
namespace: biological_process
def: "The mitotic cell cycle process in which the controlled breakdown of the nuclear pores occurs during open or closed mitosis." [PMID:32848252]
synonym: "nuclear pore complex disassembly during mitosis" EXACT []
is_a: GO:0007077 ! mitotic nuclear membrane disassembly
intersection_of: GO:0022607 ! cellular component assembly
intersection_of: part_of GO:0000278 ! mitotic cell cycle
intersection_of: results_in_disassembly_of GO:0005643 ! nuclear pore
relationship: part_of GO:0006998 ! nuclear envelope organization
property_value: term_tracker_item https://github.com/geneontology/go-ontology/issues/19947 xsd:anyURI

Note: this dodgy logical def will also show up in a query for all equiv axioms that don't match a pattern, so it's nice to know we are growing robust mechanisms for catching this sort of thing

Issue #22555, opened 2021-12-13
Inconsistency transcription inputs/outputs and GO-CAM


pgaudet:
Hello, 

In GO-CAM we model transcription as 
- MF: DNA binding transcription factor activity 
and 
- BP: 'regulation of transcription by RNA polymerase II' 

both use **'has input'**  and the gene product. 

However this is inconsistent with the ontology, where the few terms that do have a logical definition use 'has output' - for eg 

ncRNA transcription: 
'transcription, DNA-templated'
 and ('has output' some ncRNA)

This is also pretty much consistent with the design pattern of biosynthetic process, because we should even use 'has primary output'. 

Should we change the LD of transcription terms to use 'has primary output'? and in this case, we need to change the GO-CAM models to also use 'has primary output'? 

@thomaspd @cmungall @vanaukenk @balhoff 

Thanks, Pascale

ValWood:
tagging myself on this discussion @ValWood

pgaudet:
Actually, @balhoff and @cmungall say that this is not an inconsistency, as the ontology and GO-CAMs are not describing the same entities.

pgaudet:
*. **lncRNA transcription**
'transcription, DNA-templated'
and 'has primary output' some lncRNA

pgaudet:
By organelle:

* **plastid transcription**
'transcription, DNA-templated'
 and 'has primary output' some transcript
 and 'occurs in' some plastid


* **plastid rRNA transcription**
'plastid transcription'
 and 'has primary output' some rRNA

* **mitochondrial transcription**
'transcription, DNA-templated'
 and 'has primary output' some transcript
 and 'occurs in' some mitochondrion

+ subclass: part of some mitochondrial gene expression

* **mitochondrial rRNA transcription**
* 'mitochondrial transcription'
 and 'has primary output' some rRNA

ValWood:
>Actually, @balhoff and @cmungall say that this is not an inconsistency, as the ontology and GO-CAMs are not describing the same entities.

Can somebody explain these differences on a future call?

PR #22579, opened 2021-12-14
added ld for #22555

pgaudet:


cmungall:
I would like to revisit the use of process_has_causal_agent, but we can merge and iterate on this

pgaudet:
OK thanks!

Le mar. 14 déc. 2021 à 5:24 PM, Chris Mungall ***@***.***> a
écrit :

Issue #16524, opened 2018-10-17
ELK 0.5 shows GO has a large number of logical incompatibilities with property ranges


balhoff:
The latest (unreleased) ELK snapshot (obtained from https://github.com/liveontologies/elk-reasoner/wiki/GettingElk) provides reasoning on object property ranges. Classifying go-plus with ELK 0.5 shows thousands of unsatisfiable classes. Lots of these are caused by [Uberon](https://github.com/obophenotype/uberon/issues/1455), but if you delete the Uberon import you can see ~2000 remaining problems.

A quick look suggests that many of these involve regulation of qualities, e.g. [regulation of biological quality](http://purl.obolibrary.org/obo/GO_0065008). The object of 'regulates' needs to be an 'occurrent', while qualities are 'continuants'.

cmungall:
I think we need to press on ASAP with a solution, either generalizing range to a union or adding the RoQ relations

ukemi:
What these terms really mean is that we are regulating a process that results in the 'expression' of the biological quality.

ukemi:
for homeostasis we have regulates_levels_of, a sibling of regulates. Perhaps this is a route to investigate?

cmungall:
See also https://docs.google.com/document/d/1xdFtvn5Gxege4KWBrF-gwbihytY2j8qVH2CZ6_0Yluo/edit "Defining GO relations using PATO, bridging to phenotypes and LEGO"

vanaukenk:
From an annotation POV, I wonder how many annotations to 'regulation of biology quality' terms reflect phenotypes described in the paper without detailed understanding of the mechanism by which that phenotype arises.
In such cases, if we want to keep the annotations, would we be better off using a top-level relation like 'acts upstream of or within' and a BP term that represents the underlying process, even if that is a high-level term?
An example that comes to mind is 'regulation of brood size' that perhaps could instead be annotated as GP 'acts upstream of or within, positive (or negative) effect' 'reproduction'. 
We are trying to move towards an annotation practice where use of a 'regulation' term implies that the mechanism is known.  In light of that, maybe some of the higher level terms under 'regulation of biological quality' should have a 'Do Not Annotate' flag on them so that we stay focused on annotating to the underlying processes that influence that quality.

ukemi:
But something like regulation blood pressure strikes me being a bonafide process and a legitimate grouping term.

dosumis:
'regulation of membrane potential' is too

[...]

PR #22593, opened 2021-12-17
Work in progress eliminating property range problems.

balhoff:
Using three temporary relations which will be proposed to RO. I will replace the temporary relations before removing draft status.

See #16524.


balhoff:
Also waiting on https://github.com/obophenotype/uberon/pull/2219 and https://github.com/obophenotype/uberon/issues/2218.

PR #22595, opened 2021-12-18
Issue 22574 platelet rolling and its regulation terms added

raymond91125:


slaulederkind:
The new terms look good to me.  Thanks Raymond.

PR #22642, opened 2022-01-11
WIP changed relation for transporters

pgaudet:


cmungall:
This seems the salient line in the error output

` :: ERROR: ID-mentioned-twice:: has_input`

still cryptic but maybe you did a search and replace on the relation declaration itself?

cmungall:
Once the error is resolved we can see the diff

For me the safest SOP when replacing specific relations with general ones is to first have guarantees that all genus terms in DPs that use the generic relation are disjoint

However, perhaps we should just proceed and fix errors as we see them. But this requires we carefully check the reasoner diffs. Otherwise e.g. we will find that a very deep latticed term is inferred to be X transport because it traces up some path to a catabolism term for X for input.

PR #22648, opened 2022-01-11
WIP create disjoint axiom fixes #22647

### LINKED ISSUE #22647: Create disjoint axiom between GO:0008152 metabolic process and GO:0051179 localization

pgaudet:


pgaudet:
It looks like annotation of a single gene to disjoint terms is allowed (or not picked up as an error): 
http://noctua.geneontology.org/editor/graph/gomodel:61e0e55600000395

@vanaukenk @cmungall @ukemi 
Is this the expected behavior?

cmungall:
This is expected. We have to allow for situations like this, e.g. the
"same" gene can have different activities in different contexts

(the way to thin of this is that it is different instances of the gene
product in different contexts)

We would need matrix rules to catch this, no different than standard
annotations

remember, when we say 2 classes are disjoint, it means they share no
instances - it doesn't mean anything about genes per se

If you want to force an inconsistency here, this should work:

make a has-input from the ssDNA node to an instance of CHEBI:4705 !
double-stranded DNA

it should autoclassify the node as dsDNA binding

PR #22700, opened 2022-01-19
Include narrowMatch in check for only one usage of a RHEA xref.

### LINKED ISSUE #20882: Fix unicity check on Rhea xrefs
Hello, 

@balhoff has recently implemented a check that allows RHEA xrefs to only be used once in the ontology. 
We want to make the following changes: 
1. Add skos xref type = EXACT for all current Rheas
2. Fix the check so that EXACT and NARROW xrefs can only be used in a single GO term
3. Allow more than one use of a xref for BROAD xrefs

Thanks, Pascale

balhoff:
Fixes #20882.


balhoff:
This PR will fail until the xrefs in https://github.com/geneontology/go-ontology/issues/20882#issuecomment-1016634081 are fixed.

pgaudet:
Can this be merged now?  @balhoff

PR #22702, opened 2022-01-19
Added term positive regulation of ferroptosis. Fixes #22674

### LINKED ISSUE #22674: NTR: positive regulation of ferroptosis
Please provide as much information as you can: 

* **Suggested term label:**

positive regulation of ferroptosis

* **Definition (free text)**
* 
Any process that activates or increases the frequency, rate or extent of ferroptosis

* **Reference, in format PMID:#######**

PMID: 32015325

* **Gene product name and ID to be annotated to this term**

I'm not a biologist (so no expert!) but the attached review does reference more gene products associated with the mechanism (currently there are just 2 gene products associated for negative regulation), erastin (small molecule) is known the induce ferroptosis and is likely to be interacting with one or more voltage dependent ion channels (VDAC(s)) so could possibly be VDAC1/VDAC2/VDAC3

* **Parent term(s)**

GO:0110075

* **Children terms (if applicable)** Should any existing terms that should be moved underneath this new proposed term?

N/A

* **Synonyms (please specify, EXACT, BROAD, NARROW or RELATED)**

* **Cross-references**
* For enzymes, please provide RHEA and/or EC numbers. 
* Can also provide MetaCyc, KEGG, Wikipedia, and other links.

https://en.wikipedia.org/wiki/Ferroptosis 

* **Any other information**

I'm asking for the term as if you have negative regulation of a process then surely there must be +ve regulation as well.
Note: have emailed HUGO to ask if AIFM2 will be re-named FSP1 as the nature article suggests

raymond91125:


pgaudet:
please merge

PR #23052, opened 2022-03-21
Update CHEBI import.

balhoff:


balhoff:
TODO: find out why the biochebi GCIs change from 'transports or maintains localization of' to 'regulates levels of'.

PR #23164, opened 2022-04-07
Issue 23163

pgaudet:


pgaudet:
Need to update the taxon constraints - see https://github.com/geneontology/go-ontology/pull/23165

Issue #23431, opened 2022-05-31
Ontology no longer builds due to unsatisfiable classes in the ontology


kltm:
From about May 30th at midnight, the full ontology build fails with the following error:
```
09:38:23  ROBOT_JAVA_ARGS=-Xmx48G robot --catalog catalog-v001.xml merge -i extensions/go-lego-edit.ofn --collapse-import-closure true --include-annotations false reason -r ELK --exclude-duplicate-axioms true --remove-redundant-subclass-axioms true --exclude-tautologies structural relax reduce --named-classes-only true annotate -O http://purl.obolibrary.org/obo/go/extensions/go-lego.owl -V http://purl.obolibrary.org/obo/go/releases/2022-05-31/extensions/go-lego.owl -o extensions/go-lego.owl
09:39:44  ERROR There are 242 unsatisfiable classes in the ontology.
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_1902060
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0080166
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0030584
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0075259
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0030582
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048658
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048657
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048530
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048656
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048655
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048654
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0048653
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0140547
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0010069
09:39:44  ERROR     unsatisfiable: http://purl.obolibrary.org/obo/GO_0010622
[...]

PR #23444, opened 2022-06-02
Temporarily pin Uberon version.

balhoff:
@kltm this should let the build proceed.


kltm:
For: https://github.com/geneontology/go-ontology/issues/23431

PR #23767, opened 2022-07-22
Run QC on MacOS for more memory

balhoff:


balhoff:
This needs more tweaks than I realized (e.g. https://github.com/obophenotype/uberon/blob/master/.github/workflows/diff.yml) Holding off for now.

Issue #23865, opened 2022-08-19
Ontology has trouble building in some environments due to chebi-download collision


kltm:
Recently, the ontology has trouble building in some environments due to an apparent chebi-download collision:

```
00:49:38   46500K .........                                             100%  198M=4.2s
00:49:38
00:49:38  2022-08-18 07:49:38 (10.7 MB/s) - ‘mirror/chebi-download.owl.tmp.gz’ saved [47625343/47625343]
00:49:38
00:49:39  gzip: mirror/chebi-download.owl already exists;	not overwritten
00:49:39
00:49:39  real	0m5.332s
```
This seems to be able to occur even after the filesystem has been purged (unlikely to be collision from previous build).

As well, this seems to crop up a lot in the GO build pipeline since https://github.com/geneontology/pipeline/issues/296

Tagging @balhoff

kltm:
This section is interestingly hardwired, apparently for obo at some point?
https://github.com/geneontology/go-ontology/blob/dc62339ab27f3c70aaba84fc9d824e05e8870d9d/src/ontology/Makefile#L410-L414

kltm:
I've been unable to replicate using commands in the Makefile, but can replicate running 
`root@moiraine:/tools/go-ontology/src/ontology# wget --no-check-certificate http://purl.obolibrary.org/obo/chebi.owl.gz -O mirror/chebi-download.owl.tmp.gz && cp mirror/chebi-download.owl.tmp.gz mirror/chebi-download.owl.gz && gunzip mirror/chebi-download.owl.gz && touch mirror/chebi-download.owl`
multiple times (or after building with the make commands).
I wonder if there could somehow be a race or multiple builds, but we're not using parallelization...
It seems like line 410-414 should be unnecessary if we're not using OBO, as the "normal" mechanism could be used.

kltm:
This will naturally check with `snapshot` over the weekend; the workspaces have been cleared.

ValWood:
We have experienced a similar  issue with FYPO,  I believe it is an EBI ftp site issue. @kimrutherford  has reported it a number of times, and there has been some discussion about it on the phenotype-ontologies slack. If it is the same thing , I advise you also report to the EBI. Out ticket was closed as a "known issue" ???

kltm:
This has cleared.

PR #23867, opened 2022-08-19
Force gunzip output for CHEBI unzipping.

balhoff:


kltm:
@balhoff helping to clear issues around https://github.com/geneontology/go-ontology/issues/23865

PR #24039, opened 2022-09-20
Logical definition fixes based on ‘bounding membrane of organelle’ pattern.

balhoff:
These logical definitions are proposed via decomposing term names into DOS-DP pattern matches.

An additional one that could be changed is [organelle membrane](http://purl.obolibrary.org/obo/GO_0031090). However it's not clear to me from the textual definition if it is intended to be the bounding membrane of an organelle, or any membrane in an organelle.


balhoff:
@pgaudet @cmungall could you please review? Something I'm not sure of is whether the pattern applies exactly in the cases where the membrane is part of an "envelope". Is it still a `bounding layer of`?

balhoff:
This PR is currently causing some inferred equivalences:

ERROR Only equivalent classes that have been asserted are allowed. Inferred equivalencies are forbidden.
ERROR Equivalence: <http://purl.obolibrary.org/obo/GO_0033113> == <http://purl.obolibrary.org/obo/GO_0036013>
ERROR Equivalence: <http://purl.obolibrary.org/obo/GO_0070112> == <http://purl.obolibrary.org/obo/GO_0070114>
ERROR Equivalence: <http://purl.obolibrary.org/obo/GO_0042170> == <http://purl.obolibrary.org/obo/GO_0009527>
ERROR Equivalence: <http://purl.obolibrary.org/obo/GO_0009668> == <http://purl.obolibrary.org/obo/GO_0009666>
ERROR Equivalence: <http://purl.obolibrary.org/obo/GO_0031969> == <http://purl.obolibrary.org/obo/GO_0009707>
ERROR Equivalence: <http://purl.obolibrary.org/obo/GO_0046862> == <http://purl.obolibrary.org/obo/GO_0031900>

balhoff:
> Something I'm not sure of is whether the pattern applies exactly in the cases where the membrane is part of an "envelope". Is it still a `bounding layer of`?

Oh I see, in all those cases there is already an "outer membrane" with that definition.

src/ontology/go-edit.obo
```diff
@@ -187472,8 +187473,8 @@ namespace: cellular_component
 def: "Either of the two lipid bilayers surrounding a Cvt vesicle, a vesicle that functions in the cytoplasm-to-vacuole targeting (Cvt) pathway." [GOC:ecd, PMID:20065092]
 synonym: "cytoplasm to vacuole targeting vesicle membrane" EXACT []
 synonym: "cytoplasm-to-vacuole targeting vesicle membrane" EXACT []
-is_a: GO:0030659 ! cytoplasmic vesicle membrane
-relationship: part_of GO:0033107 ! Cvt vesicle
```
cmungall, line 187476: this will definitely return in the release version, yes?

(it's a sort of non-standard relax+roll-up, I don't recall if we have a specific procedure for doing this)

cmungall (commented):
looks ok provided we infer the rollups to part-of

src/ontology/go-edit.obo
balhoff: We do! We materialize 'part of'.

cmungall (approved):

balhoff:
Before merging this, I need to revise the definitions for terms that have "outer" membrane counterparts.

PR #24202, opened 2022-10-17
Use https instead of ftp for EBI files

### LINKED ISSUE #24195: Switch EBI URLs in the metadata from ftp to http(s) schema
We are having occasional ontology build failures in the pipeline related to the EBI FTP not being available. As with https://github.com/geneontology/pipeline/issues/304, we should switch over from FTP to HTTP(S).

Tagging @balhoff

balhoff:
Fixes #24195.

@kltm does this cover everything you had in mind?


kltm (approved):
Sure does!
Thank you @balhoff

Issue #17613, opened 2019-07-12
leucocyte terms for taxon constraint


hattrill:
Could you put a taxon constrain on:

GO:0002443 | leukocyte mediated immunity
GO:0045321 | leukocyte activation
GO:0050900 | leukocyte migration

I think that the appropriate taxid is Protostomia NCBI:txid33317

(From the CL:
hemocyte (sensu Nematoda and Protostomia) CL:0000387 are the immune cells of inverts (rather than leukocyte CL:0000738 which is a vertebrate thing))

(I haven't exhaustively gone through the GO to look for leukocyte terms, just these nodes looked like good bets for what I was looking at.)

pgaudet:
Taxon constraints should exist !! We just need to get them in the reports.....

ukemi:
Would it make more sense to have these constraints in CL on the cell type and then create valid property chains to inherit them in GO? That way we would be sure not to miss any. 
I don't see any declared taxon constraints in cl.owl.
It would be interesting to devise a hierarchy of taxon restraints  UBERON-CL-GO
@balhoff?

pgaudet:
> Would it make more sense to have these constraints in CL on the cell type 

I thought this was already what was happening.

balhoff:
Like @ukemi I didn't see any taxon constraints in CL. I think @dosumis and @matentzn have been working on CL lately. Any plans for this?

dosumis:
No plans right now - but we should definitely add them.  In the first instance, we should add these terms where the names make this explicit (e.g. hemocyte (sensu Nematoda and Protostomia)) IIRC there are some explicitly arthropod terms too.  Do you have a list of cases where you need this for GO right now?  If so, could you put them on a CL ticket?  Even better if you can do a pull request.

ukemi:
Yes. Some could be inherited from Uberon, a good start but they won't be specific enough for all classes. It just seems like a better strategy to add them to the elemental unit than for GO to have to add them to every term that involves a certain cell type.
But I guess the fix for now is to do that.

krchristie:
taxon constraint was added in CL on 9/6/19 so should appear in next release and then will be available

krchristie:
@hattrill - The CL term for `leucocyte` now has a taxon constraint on it. Has this solved your issue?

I think plans are in the works to allow GO editors to see constraints inherited from other ontologies, but it isn't in place yet.

hattrill:
Thanks.
[...]

PR #24203, opened 2022-10-17
Added taxon constraints to leukocyte terms. #17613

raymond91125:


raymond91125:
@balhoff Jim, do you know what's blocking?
Is this it? "ls: cannot access 'subsets/*.owl': No such file or directory"
But I don't know where to find the owl files anyway. thanks.

balhoff:
@raymond91125 I restarted the check—I think it was just a github glitch. As long as the "Required" one passes you can merge.

Issue #24222, opened 2022-10-19
Hemeostasis: regulation of liquid surface tension: merge into surfactant homeostasis


pgaudet:
Hello, 

For the BP refactoring, we will split cellular and organism-level processes, as described in the [2022 bp refactoring plan - top level](https://docs.google.com/document/d/1k8yuWTMSkYVTnt2hRbrPVH8Ud3gDwC5IrGu4PqHkKds/edit). 

GO:0050828 regulation of liquid surface tension is a direct child of [GO:0042592 homeostatic process](http://amigo.geneontology.org/amigo/term/GO:0042592#display-lineage-tab), which will be obsoleted, see  #24218
This term will be merged and end up a child of 'GO:0048871 multicellular organismal homeostasis'

 - [x] merge into GO:0050828 regulation of liquid surface tension into GO:0043129 surfactant homeostasis 
 - [x] obsolete GO:0101029 negative regulation of liquid surface tension & GO:0101028 positive regulation of liquid surface tension (no annotations)
 - [x] change definition to "Any process involved in the maintenance of a steady-state level of a surface-active agent that maintains the surface tension of a liquid."
 - [x] add reference: PMID:23708874
 - [x] GO:0043129 surfactant homeostasis -> move under 'GO:0048871 multicellular organismal homeostasis'

Thanks, Pascale

PR #24233, opened 2022-10-20
Merged regulation of liquid surface tension into surfactant homeostas…

raymond91125:
…is. #24222


raymond91125:
@balhoff CI failed twice, not sure why. Does it have something to do with the removal of dcterms:conformsTo? Thanks.

balhoff:
@raymond91125 how did that happen that `go-pattern-conformance.ttl` was edited? It's autogenerated and shouldn't be changed manually.

balhoff:
I guess merging terms results in some tricky Protege behavior with the external file. I would recommend just not committing any changes that appear in files outside of go-edit.obo. Could you redo the PR and not include the changes to that file?

raymond91125:
I think protege did that. I was merging a regulation term into a 'non-regulation' term.

PR #24234, opened 2022-10-20
Merged regulation of liquid surface tension into surfactant homeostas…

raymond91125:
…is. #24222


raymond91125:
FAIL Rule ../sparql/replacedby-obsolete-violation.sparql: 1 violation(s)
class,replacement
http://purl.obolibrary.org/obo/GO_0019214,http://purl.obolibrary.org/obo/GO_0050828
In Protege, searching for the term label "regulation of liquid surface tension" does not find the replaced by usage. Must search by ID.

balhoff:
@raymond91125 I think the issue is that http://purl.obolibrary.org/obo/GO_0019214 has a `consider` link to http://purl.obolibrary.org/obo/GO_0050828, but http://purl.obolibrary.org/obo/GO_0050828 is obsolete. So that needs to be changed to GO:0043129.

balhoff:
Oh I think you already figured that out, let me know if anything else is going wrong.

raymond91125:
Thanks, @balhoff. Yes, I finally saw those lines. Going blind.

Issue #24248, opened 2022-10-21
Advice: development terms related to the plant endodermis


tberardini:
Review of recent QC reports showed that there are some plant genes annotated to %endoderm% biological process terms. The two I was looking at were CACAO annotations to 'endodermal cell fate specification , GO:0001714'. It looks like most (all?) such terms are linked (transitively?) to the CL term for endoderm. Reading the def of the above term doesn't give the clue that it is animal specific. 

> The cell fate determination process that results in a cell becoming capable of differentiating autonomously into an endoderm cell in an environment that is neutral with respect to the developmental pathway; upon specification, the cell fate can be reversed. Source: GOC:go_curators)

There are several plant anatomy terms that share this root word and GO BP terms exist that look from the surface like they would apply to plants as well.

https://browser.planteome.org/amigo/search/ontology?q=endoderm
http://amigo.geneontology.org/amigo/search/ontology?q=endoderm

Example paper:
SCARECROW-LIKE23 and SCARECROW jointly specify endodermal cell fate but distinctly control SHORT-ROOT movement
http://doi.org/10.1111/tpj.13038

How would you like to handle this situation?
Should we request terms that are specific to the plant endodermis like 'plant endodermal cell fate specification?' Does it make sense to rename the existing generic sounding terms to something like 'animal endodermal cell fate specification?'

tberardini:
@lreiser Looping you in.

raymond91125:
It appears that endoderm (animal embryo) and endodermis (plant ground tissue) were meant to indicate the *inner* cell layer in different anatomical contexts. Thus it would not make sense to mix them together. As it was pointed out, the existing GO terms are defined by the animal specific term CL:00000223 endodermal cell.
Is the word "endomermis" restricted to plants? If so:
I think it would be best if GO have endodermis-specific terms that are logically defined with a CL or Uberon term (perhaps "cell of endodermis", "endodermis, respectively). The CL and Uberon terms would be defined as plant-specific and the GO terms will inherit the taxon constraints.
In GO, we could have terms that are, e.g. "endodermal cell fate specification, plant ground tissue" and they are 'siblings' (not child) of "endodermal cell fate specification". @pgaudet Are terms like that OK still?

pgaudet:
Thanks for looking into this @raymond91125 

This does seem reasonable, the animal and plant processes must be very different. 

[...]

PR #24257, opened 2022-10-25
Created plant endodermal cell fate specification & plant endodermal c…

raymond91125:
…ell differentiation. #24248


tberardini:
Looks good, @raymond91125.  Please go ahead with pushing into the master version.

Issue #17084, opened 2019-03-28
Is hormone secretion intended only for endocrine, or includes other types?


cmungall:
[this issue is for not discussion of CHEBI roles. It pertains entirely to the usage of the term "hormone"]

We have:

```
id: GO:0046879
name: hormone secretion
namespace: biological_process
def: "The regulated release of hormones, substances with a specific regulatory effect on a particular organ or group of cells." [ISBN:0198506732]
is_a: GO:0008150 ! biological_process
is_a: GO:0009914 ! hormone transport
is_a: GO:0023061 ! signal release
```

There is only one subclass that indicates mode of dispersal:

```
id: GO:0060986
name: endocrine hormone secretion
namespace: biological_process
def: "The regulated release of a hormone into the circulatory system." 
is_a: GO:0046879 ! hormone secretion
relationship: part_of GO:0050886 ! endocrine process
```

There is no subclasses for autocrine, paracrine, etc.

I think GO:0060986 is fake precision. I think people read "hormone secretion" as being endocrine. 

If you query with GO:0060986 then you miss insulin secretion! There are only 87 mouse annotations to GO:0060986, as opposed to GO:0046879 for hormone signaling. From this we would conclude that endocrine hormone signaling was a rarer form of hormone signaling.

Also, not sure GO:0060986 def is correct. The def includes exocrine system. I think it intends to say releases *directly* into ducts.

Do we really need this level of precision at all? Mechanistically is there a difference in the gene product action between endocrine/exocrine/paracrine/autocrine?

If not, then we can simplify and merge

If there is, then we will have to do work to (a) properly clarify that the parent term encompasses different subtypes and include a do_not_annotate (since you always know the subtype) (b) introduce children not just for endocrine (c) reannotate so we do not have false precision and unnevenness

i haven't looked at the other usages of hormone

cmungall:
I suppose the rationale for not putting insulin secretion under endocrine is the fact worms do not have a circulatory system, and we have worm annotations http://amigo.geneontology.org/amigo/reference/PMID:17289575

But this reinforces my point: do we even need this level of distinction?

cmungall:
Aside:

We have nothing linking endocrine process to uberon or ncbitaxon hence to TCs, we have one worm annotation to an endocrine process (angiotensin maturation)

http://amigo.geneontology.org/amigo/gene_product/WB:WBGene00044150

This comes from an interpro2go:
http://www.ebi.ac.uk/interpro/entry/IPR033508

[...]

PR #24473, opened 2022-11-29
Added logical definitions for hormone secretion and endocrine hormone…

raymond91125:
… secretion. #17084


src/ontology/go-edit.obo
```diff
@@ -278773,6 +278773,8 @@ namespace: biological_process
 def: "The regulated release of hormones, substances with a specific regulatory effect on a particular organ or group of cells." [ISBN:0198506732]
 is_a: GO:0009914 ! hormone transport
 is_a: GO:0023061 ! signal release
+intersection_of: GO:0046903 ! secretion
+intersection_of: has_primary_input CHEBI:24621 ! hormone
```
dosumis, line 278777: @cmungall - should this be using the GOCHE term for molecule with hormone role?

balhoff: I think so. Sadly the checks aren't failing.

raymond91125: When and how should we use GOCHE vs CHEBI?

balhoff: Any time you need to use a CHEBI role in a place where a material entity is needed. In CHEBI, 'hormone' is classified under role.

raymond91125: I see. Thanks.

PR #24498, opened 2022-12-03
Standardize cross-reference to TXPO with CURIE

cthoyt:
A cross-reference to a TXPO term appeared as an OBO PURL rather than a CURIE. This PR fixes it.


balhoff:
Thanks for the fix @cthoyt! The only hold-up merging is I'm trying to decide if I need to add this prefix to https://github.com/geneontology/go-site/blob/master/metadata/db-xrefs.yaml. Ideally I think these things should natively be IRIs but we have some issues with that in our OBO source format.

PR #24671, opened 2023-01-10
Workflow that runs `ontobot-change-agent` when relevant issue is created/edited.

hrshdhgd:
If an issue is made relevant to editing the ontology file, this workflow will run to trigger a PR. This includes 1 file:
1. The workflow YAML file itself


cmungall (approved):
Thanks!

Ontology team: this adds functionality that @balhoff and I can demonstrate on a future call

PR #24694, opened 2023-01-12
Standardize `GOC:cjm` references to use ORCID

cthoyt:
As a follow-up to https://github.com/cthoyt/orcidio/issues/5, this PR demonstrates updating the provenance in various places to use ORCID identifiers as a more actionable reference for @cmungall instead of `GOC:cjm`

As part of this PR, I also issued https://github.com/geneontology/go-site/pull/1956, which adds `orcid` to the GO registry because the following code relies on the live deploy of this file for validation of xrefs:

https://github.com/geneontology/go-ontology/blob/4e61ae2c6956428524555d535c578dfabbc30d0d/src/ontology/Makefile#L152-L159


cthoyt:
@kltm I think this is ready for review. It appears a CI command has failed due to some issue with it interacting with GitHub's API, so I'm not sure if it has any relevant feedback or not

balhoff:
@cthoyt just wanted to note the reason why I have delayed merging this. There are a number of issues with the OBO format OWL translation which we're working on, including mapping CURIE values to actual IRI resources rather than strings (https://github.com/owlcs/owlapi/pull/1072). As it stands, all these `orcid` CURIEs will become strings when parsed into OWL, thwarting interoperability with records in your ORCIDIO ontology. So I wanted to try to get those updates into OWLAPI and then roll this out for all GO contributors.

kltm:
I think we should also shoot for cleanly clearing tests for any merge, regardless of reason. If a test cannot consistently run, it should be removed.

balhoff:
Only the ontology_qc test is required for merge. The others are informational. The last one fails for PRs outside the org (probably something I can fix with better configuration).

cthoyt:
Thanks for the note @balhoff. If there’s anything I can do to support you, please let me know.

cthoyt:
@balhoff if I change the ORCID CURIEs to URIs (and resolve conflicts), can this get merged?

kltm:
@balhoff Would this have any consequences for the Noctua data flow and client expectations? If so, we'll want to take and changes slowly.

Issue #24726, opened 2023-01-17
Obsoletion notice: GO:0030104 water homeostasis


pgaudet:
Dear all, 

The proposal has been made to obsolete GO:0030104 water homeostasis. The reason for obsoletion is that this term does not specify whether the process is intracellular or organism-level. Consider: 
*  GO:0009992 intracellular water homeostasis
* GO:0050891 multicellular organismal-level water homeostasis 
* GO:0003091 renal water homeostasis

There are 30 annotations to this term see https://github.com/geneontology/go-annotation/issues/4407
Group | Number of annotations
-- | --
dictyBase | 1
FlyBase | 5
MGI | 5
TAIR | 3
UniProt | 8
WB | 4
ZFIN | 4

There are no mappings, this term is not in any subsets. 

You can comment on the ticket: 

Thanks, Pascale

pfey03:
Dicty done

raymond91125:
When I tried to obsolete GO:0030104 water homeostasis (Protege 5.5.0 Build beta-8, Edit->Deprecate entry...), I get errors for pull requests. I noticed that one strange thing is that although in Protege I see owl:deprecated true, there is no is_obsolete: true in the go-edit.obo file Protege saved. @pgaudet @balhoff Do you know why? Should I manually add that line?

```
diff --git a/src/ontology/go-edit.obo b/src/ontology/go-edit.obo
index cac26669..291d5910 100644
--- a/src/ontology/go-edit.obo
+++ b/src/ontology/go-edit.obo
@@ -25458,8 +25458,6 @@ def: "Renal process involved in the maintenance of an internal steady state of w
 synonym: "water homeostasis by the renal system" EXACT []
 is_a: GO:0003014 ! renal system process
 is_a: GO:0050891 ! multicellular organismal-level water homeostasis
-intersection_of: GO:0030104 ! water homeostasis
-intersection_of: occurs_in UBERON:0001008 ! renal system
 
 [Term]
 id: GO:0003092
@@ -137264,6 +137262,11 @@ def: "Synthesis and ordering of the envelope of pronuclei." [GOC:ems]
 is_a: GO:0006998 ! nuclear envelope organization
 relationship: part_of GO:0007338 ! single fertilization
 
+[Term]
+id: GO:0018987
+property_value: IAO:0000231 IAO:0000227
+is_obsolete: true
+
 [Term]
 id: GO:0018988
 name: obsolete molting cycle, protein-based cuticle
[...]

PR #24792, opened 2023-01-24
Obsoleted water homeostasis. #24726

raymond91125:


raymond91125:
mistake made. start new.

Issue #24907, opened 2023-02-13
Shouln't phagocytosis be is_a endocytosis? 


pgaudet:
GO:0006909 phagocytosis: A vesicle-mediated transport process that results in the engulfment of external particulate material by phagocytes and their delivery to the lysosome. The particles are initially contained within phagocytic vacuoles (phagosomes), which then fuse with primary lysosomes to effect digestion of the particles.

GO:0006897 endocytosis: A vesicle-mediated transport process in which cells take up external materials or membrane constituents by the invagination of a small region of the plasma membrane to form a new membrane-bounded vesicle.

but these are siblings: 

<img width="246" alt="image" src="https://user-images.githubusercontent.com/4782928/218516537-8147306d-9455-4e3b-9dfe-9869b6feca47.png">

addiehl:
I think this is a question of mechanism.  Phagocytosis does not rely on invagination of a small region of the plasma membrane, but rather a much larger region of the plasma membrane.

addiehl:
[Difference Between Endocytosis and Phagocytosis](https://www.differencebetween.com/difference-between-endocytosis-and-vs-phagocytosis/)

raymond91125:
That article also states: "Endocytosis is of three categories. Among them, phagocytosis and pinocytosis are the two most common types. Moreover, phagocytosis is a form of endocytosis."

raymond91125:
We probably should remove the "small region" part of the definition.
New definition: A vesicle-mediated transport process in which cells take up external materials or membrane constituents by the invagination of a part of the plasma membrane to form a new membrane-bounded vesicle.
Add a reference [https://en.wikipedia.org/wiki/Endocytosis].
Then make phagocytosis a sub-type of endocytosis. Pinocytosis and receptor-mediated endocytosis are already types of endocytosis.

addiehl:
I am happy with the definition change of 'endocytosis' if this change is going to be made.  I would also point out (as I thought I remembered), that 'phagocytosis' was a child of 'endocytosis' until 2018-12-25, per [QuickGO](https://www.ebi.ac.uk/QuickGO/GTerm?id=GO:0006909#term=history), when its definition was changed and it was elevated to be a sibling of 'endocytosis'.

You might want to review **location of 'phagocytosis' in the ontology** #16576 and PR #16824 and hear back from @RLovering and @ukemi prior to making these changes. It's not a good look for GO to be flippy-floppy on these basic cell biology classes.

[...]

Issue #16576, opened 2018-11-06
location of 'phagocytosis' in the ontology


RLovering:
Hi
Looking at the definitions of the following related terms I think that their relationship needs to be revised:
 GO:0006897 endocytosis
A vesicle-mediated transport process in which cells take up external materials or membrane constituents by the invagination of a small region of the plasma membrane to form a new membrane-bounded vesicle. 
>is_a GO:0006909 phagocytosis
An endocytosis process that results in the engulfment of external particulate material by phagocytes. The particles are initially contained within phagocytic vacuoles (phagosomes), which then fuse with primary lysosomes to effect digestion of the particles. 

The term GO:0006909 phagocytosis describes the process of fusion with lysosomes which is not an 'endocytic' process.

I think that  GO:0006911 phagocytosis, engulfment should be an is_a child of GO:0006897 endocytosis, rather than a grandchild (which is what it is currently). As well as remaining a_part_of GO:0006909 phagocytosis.

I think that  GO:0044248 cellular catabolic process should be the new is_a parent to GO:0006909 phagocytosis.

Thanks

Ruth

ukemi:
Hi @RLovering, 

I'm doing a bit of background reading to be sure that this is the way protists work too, but I think that eventually all phagocytic vesicles fuse with lysosomes. Most of the papers I have looked at say that phagocytosis is a type of endocytosis, but then they do go on to describe what happens to the phagosome after it is formed. So I guess the big question is do all types of endocytosis end with the formation of the endocytic vesicle or can the process extend? I don't think our definition of endocytosis excludes the other steps of phagocytosis, because it is classified as a vesicle-mediated transport process. This seems do have been done purposefully since the subsequent fusion steps are part of phagocytosis. If we agree with this, then phagocytosis is a legitimate type of endocytosis. This seems to be the view of the community. I do think there needs to be some work on this part of the ontology. It seems that there are relationships missing.

RLovering:
OK if that is how you feel. I just think that endocytosis starts with the engulfment step, and ends when the plasma membrane is separated to form a vesicle. 

[...]

PR #24908, opened 2023-02-14
Revised definition of endocytosis to cover phagocytosis. #24907

raymond91125:


addiehl:
Please consider my comment on #24907 and the prior changes to this hierarchy from 2018 (#16576 and #16824) before making this change.

raymond91125:
@addiehl After some deliberation at the ontology editor's meeting, we decided that phagocytosis should indeed is_a endocytosis.

addiehl:
Great, thanks for the update.

Issue #24901, opened 2023-02-10
Slims have no date/version, should be required


suzialeksander:
This was pointed out by a third party, but there is no version or date on the slim files.  Can we add creation dates if known, and specify files now need to at least include the GO release they're based on? Would prefer both GO release & creation date. Not sure if there's a specs doc for the headers anywhere.

pgaudet:
I think this is a pipeline task - @kltm what do you think? Can we move repos?

kltm:
@pgaudet The code for this is in the ontology build and the data for this is in the ontology, I believe this is the repo where all activity will take place.

balhoff:
@kltm @pgaudet can you point me to a file in the release that represents what this issue is about?

kltm:
@balhoff I believe an example would be
http://current.geneontology.org/ontology/subsets/goslim_yeast.obo
It looks like
http://current.geneontology.org/ontology/subsets/goslim_yeast.owl
has a (dated) version, but not a date?

pgaudet:
We should indeed sort out what the headers should look like. The yeast slim.obo for example has 

format-version: 1.2
subsetdef: chebi_ph7_3 "Rhea list of ChEBI terms representing the major species at pH 7.3."
subsetdef: gocheck_do_not_annotate "Term not to be used for direct annotation"
subsetdef: gocheck_do_not_manually_annotate "Term not to be used for direct manual annotation"
subsetdef: goslim_agr "AGR slim"
subsetdef: goslim_aspergillus "Aspergillus GO slim"
subsetdef: goslim_candida "Candida GO slim"
subsetdef: goslim_chembl "ChEMBL protein targets summary"
subsetdef: goslim_drosophila "Drosophila GO slim"
subsetdef: goslim_flybase_ribbon "FlyBase Drosophila GO ribbon slim"
subsetdef: goslim_generic "Generic GO slim"
subsetdef: goslim_metagenomics "Metagenomics GO slim"
subsetdef: goslim_mouse "Mouse GO slim"
subsetdef: goslim_pir "PIR GO slim"
subsetdef: goslim_plant "Plant GO slim"
subsetdef: goslim_pombe "Fission yeast GO slim"
subsetdef: goslim_synapse "synapse GO slim"
subsetdef: goslim_yeast "Yeast GO slim"
subsetdef: prokaryote_subset "GO subset for prokaryotes"
synonymtypedef: syngo_official_label "label approved by the SynGO project"
[...]

PR #24976, opened 2023-02-21
Explicitly list published subsets.

balhoff:
For #24901.


kltm:
@balhoff will merge and test

PR #24989, opened 2023-02-23
Vw 23 02 23 part c

ValWood:


balhoff:
@ValWood here is the error:

```
FAIL Rule ../sparql/non-anyURI-value-violation.sparql: 2 violation(s)
goTerm,property,value
http://purl.obolibrary.org/obo/GO_0000738,http://purl.obolibrary.org/obo/IAO_0000233,https://github.com/geneontology/go-ontology/issues/24075
http://purl.obolibrary.org/obo/GO_0000737,http://purl.obolibrary.org/obo/IAO_0000233,https://github.com/geneontology/go-ontology/issues/24075
```

To fix this you just need to change the datatype of your GitHub link string. I can show you how to do this on zoom if you like.

ValWood:
RIght, got it. I forgot to reset that after Protege crashed!

Issue #23910, opened 2022-08-29
Taxon constraint: GO:0035658 | Mon1-Ccz1 complex


ValWood:
Please provide as much information as you can: 

GO:0035658 | Mon1-Ccz1 complex
never in Schizosaccharomyces

* **Supporting evidence if available (e.g PMID):**

see https://github.com/geneontology/go-annotation/issues/4266

ValWood:
This one was done a while back, forgot to close

PR #25074, opened 2023-03-03
Update never_in_taxon.tsv

ValWood:
https://github.com/geneontology/go-ontology/issues/23910 Note that it a while since I did one of these. There was a blank line at the bottom of the file which I preserved (although it seemed odd) so if there is a problem, that might be it.


ValWood:
It failed, I have no idea why. @balhoff @pgaudet

pgaudet:
You were missing a 5th column  for the PMID - where we put an empty tab if there is no data. 

Do you want to merge this ?

Issue #24402, opened 2022-11-22
GO:0033862 UMP kinase activity and  GO:0009041 uridylate kinase activity are the same activity


pgaudet:
See https://github.com/geneontology/go-annotation/issues/3429

These should be either merged or obsoleted with replaced_by 

See also https://www.rhea-db.org/rhea?query=2.7.4.14+

pfey03:
@pgaudet let me know what's the decision, we have 1 EXP UMP kinase activity annotation

pgaudet:
We need to obsolete + replace by

Thanks for the ping!

raymond91125:
GO:0009041 uridylate kinase activity is a grouping term (has no direct annotations) for GO:0033862 UMP kinase activity and GO:0120136 dUMP kinase activity and both have EXP annotations. Thus, I don't think it makes sense to merge just one child with the parent.
https://enzyme.expasy.org/EC/2.7.4.14 uses CMP, dCMP, or UMP, which seems to be a different enzyme.
The label for GO:0009041 "Uridylate" appears to be causing the confusion. Perhaps we can change it to UMP/dUMP kinase activity.

raymond91125:
[Term]
 id: GO:0009041
-name: uridylate kinase activity
+name: UMP/dUMP kinase activity
 namespace: molecular_function
 def: "Catalysis of the reaction: ATP + (d)UMP = ADP + (d)UDP." [GOC:go_curators]
-xref: EC:2.7.4.22
 is_a: GO:0016776 ! phosphotransferase activity, phosphate group as acceptor
 is_a: GO:0050145 ! nucleoside monophosphate kinase activity
+property_value: term_tracker_item "https://github.com/geneontology/go-ontology/issues/24402" xsd:anyURI

raymond91125:
@pgaudet do you agree with this solution? There would be no need of moving annotations.

pgaudet:
yes that works!

raymond91125:
thanks.

PR #25146, opened 2023-03-17
Renamed uridylate kinase activity. #24402

raymond91125:


pgaudet:
I think this can be merged?

Issue #19120, opened 2020-03-20
Create "materialized" ontology for AmiGO /GOlr (go-gaf)


kltm:
It would be nice to look at the current AmiGO load

"http://skyhook.berkeleybop.org/master/ontology/extensions/go-gaf.owl",
"http://skyhook.berkeleybop.org/master/ontology/extensions/gorel.owl",
"http://skyhook.berkeleybop.org/master/ontology/extensions/go-modules-annotations.owl",
"http://skyhook.berkeleybop.org/master/ontology/extensions/go-taxon-subsets.owl",
"http://purl.obolibrary.org/obo/eco/eco-basic.owl",
"http://purl.obolibrary.org/obo/ncbitaxon/subsets/taxslim.owl",
"http://purl.obolibrary.org/obo/cl/cl-basic.owl",
"http://purl.obolibrary.org/obo/pato.owl",
"http://purl.obolibrary.org/obo/po.owl",
"http://purl.obolibrary.org/obo/chebi.owl",
"http://purl.obolibrary.org/obo/uberon/basic.owl",
"http://purl.obolibrary.org/obo/wbbt.owl"

It would be good to do for AmiGO what we've done for Minerva and the Noctua AC with go-lego.owl: make a single "materialized" ontology that can be easily reused within the pipeline (and for end users).

balhoff:
Could this be an enhancement to go-gaf? Or, what other uses does go-gaf have besides being loaded here?

kltm:
@balhoff Definitely. I'm not sure there are actually other uses--the "gaf" and "lego" parts referring to their intended data context. That said, the overall load has accrued some other ontologies. Ideally, we could trim the fat and be left with a single product, whether something new or just correcting go-gaf.

ukemi:
I would be a fan of one ontology to bind them all. I've never really been a big fan of having different ontologies for different purposes.

kltm:
@balhoff After being pinged by @dougli1sqrd , it looks like go-gaf.owl is already materialized. That means that only the rationalization between go-lego and go-gaf needs to be accomplished.

kltm:
go-gaf itself is:
http://purl.obolibrary.org/obo/go/extensions/gorel.owl
http://purl.obolibrary.org/obo/cl/cl-basic.owl
http://purl.obolibrary.org/obo/go/extensions/go-plus.owl
http://purl.obolibrary.org/obo/ncbitaxon/subsets/taxslim.owl
[...]

PR #25216, opened 2023-03-31
Pre-build ontology for AmiGO loader.

balhoff:
This allows us to check reasoning compatibility within the ontology pipeline instead of the AmiGO pipeline, and also reuses downloaded imports across ontology products.

For #19120.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -81,7 +81,7 @@ REPORTDIR = $(RELEASEDIR)/reports
 # TOP LEVEL TARGETS
 # ----------------------------------------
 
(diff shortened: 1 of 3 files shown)


kltm:
@balhoff Checking that this is purely an addition? (By my read it is.)
If so, I'll go ahead and merge ASAP for testing.

balhoff:
Mostly yes, although it tweaks the catalog a bit for go-lego and go-gaf. If an issue comes up with those, I'll make a quick fix.

kltm:
@balhoff @pgaudet Okay, let's go with this for now.

Issue #25237, opened 2023-04-05
Build failing on checks relating to obsolete classes (GO:0106360)


kltm:
The NEO and go-ontology builds are currently failing on checks relating to obsolete classes (GO_0106360):

```
08:40:27  ROBOT_JAVA_ARGS=-Xmx48G robot --catalog catalog-v001.xml verify --input reasoned.owl --queries ../sparql/multiple-class-links-violation.sparql ../sparql/obsolete-reference-violation.sparql ../sparql/missing_superclass-violation.sparql -O reports/
08:41:13  PASS Rule ../sparql/multiple-class-links-violation.sparql: 0 violation(s)
08:41:13  FAIL Rule ../sparql/obsolete-reference-violation.sparql: 1 violation(s)
08:41:13  obsolete
08:41:13  http://purl.obolibrary.org/obo/GO_0106360
08:41:13  PASS Rule ../sparql/missing_superclass-violation.sparql: 0 violation(s)
```

Tagging @balhoff @pgaudet

kltm:
Testing @balhoff 's fix.

kltm:
Seems to be fixed now.

pgaudet:
Thanks both!

PR #25238, opened 2023-04-05
Update Rhea participants.

balhoff:
For #25237.


kltm:
Taking from convo w @balhoff

Issue #25288, opened 2023-04-18
Create new subset 'term_scheduled_for_obsoletion'


pgaudet:
term_scheduled_for_obsoletion "This term will be obsoleted and should not be used for new annotations."

pgaudet:
@balhoff 

Would you please add this subset to 
GO:0065009 regulation of molecular function & all children and 
GO:0036211 protein modification process & all children

Please just create a branch, and I will edit the branch to remove the children we want to keep from that subset. 

Thanks, Pascale

pgaudet:
Done

pgaudet:
In fact we have decided to use the existing subset 'gocheck_do_not_annotate' instead of 'term_scheduled_for_obsoletion'

pgaudet:
Fixed

pgaudet:
Reverted the decision - created a new subset: gocheck_obsoletion_candidate

PR #25295, opened 2023-04-19
Add subset 'term_scheduled_for_obsoletion' to subclasses of GO:0065009 and GO:0036211.

balhoff:
See #25288.


pgaudet:
@balhoff Can we merge this ? (when checks are finished) - I suppose that will generate the list of terms in that subset here http://release.geneontology.org/2023-04-01/ontology/subsets/index.html? 

Thanks, Pascale

Issue #25314, opened 2023-04-21
cholestenone 5-alpha-reductase activity GO:0047751


sjm41:
cholestenone 5-alpha-reductase activity GO:0047751 has xref of EC:1.3.1.22, but that is for (at least nowadays) a more general reaction:

https://enzyme.expasy.org/EC/1.3.1.22
name: 3-oxo-5alpha-steroid 4-dehydrogenase (NADP(+))
alt_names: 3-oxo-5alpha-steroid 4-dehydrogenase, cholestenone 5alpha-reductase
reaction: a 3-oxo-5alpha-steroid + NADP(+) <=> a 3-oxo-Delta(4)-steroid + H(+) + NADPH

Should we remove the EC xref, or update the GO term and def to match EC? (I'd vote for the latter)

The current RHEA:24552 xref is accurate for 'cholestenone 5-alpha-reductase activity'.  If we make the GO term more general and match EC:1.3.1.22, then the RHEA xref should change to the more general [RHEA:54384](https://www.rhea-db.org/rhea/54384)

id: GO:0047751
name: cholestenone 5-alpha-reductase activity
namespace: molecular_function
def: "Catalysis of the reaction: 5alpha-cholestan-3-one + NADP+ = cholest-4-en-3-one + H+ + NADPH." [EC:1.3.1.22, RHEA:24552]
synonym: "3-oxo-5alpha-steroid:NADP+ delta4-oxidoreductase activity" RELATED [EC:1.3.1.22]
synonym: "3-oxosteroid 5alpha-reductase activity" RELATED [EC:1.3.1.22]
synonym: "4-ene-3-oxosteroid 5alpha-reductase activity" RELATED [EC:1.3.1.22]
synonym: "4-ene-5alpha-reductase activity" RELATED [EC:1.3.1.22]
synonym: "cholest-4-en-3-one 5alpha-reductase activity" RELATED [EC:1.3.1.22]
synonym: "cholestenone 5a-reductase activity" EXACT []
synonym: "cholestenone 5alpha-reductase activity" RELATED [EC:1.3.1.22]
synonym: "delta4-3-ketosteroid 5alpha-oxidoreductase activity" RELATED [EC:1.3.1.22]
synonym: "reduced nicotinamide adenine dinucleotide phosphate:Delta4-3-ketosteroid 5alpha-oxidoreductase activity" RELATED [EC:1.3.1.22]
[...]

PR #25324, opened 2023-04-27
Added back RHEA. #25314

raymond91125:


raymond91125:
@balhoff I don't understand this error,  could you help? Thanks.
FAIL Rule ../sparql/one-to-one-xrefs-by-subject-violation.sparql: 1 violation(s)
c,one_to_one_prefix
http://purl.obolibrary.org/obo/GO_0047751,RHEA:

balhoff:
@raymond91125 the rule checks that there is not more than one exact RHEA xref for a given term. But looking at your diff, I don't see more than one, so I'm confused also.

balhoff:
@raymond91125 I see the issue, we have an automatic migration of RHEA xrefs from obsolete terms to their replaced_by terms. So during the ontology build the xref from GO:0050213 gets moved to this term. If you think it makes sense, I would remove the RHEA xref from GO:0050213.

raymond91125:
Ah, was wondering about that. Thanks.

Issue #25321, opened 2023-04-27
Updated definitions for three radial spoke terms


krchristie:
These three existing radial spoke terms need updates to their definitions and relationships. I worked in Protege 5.6.1, but don't have my authorization set up to commit to GitHub right now.

**old stanza**

> [Term]
> id: GO:0001534
> name: radial spoke
> namespace: cellular_component
> def: "Protein complex that links the outer microtubule doublet of the ciliary or flagellum axoneme with the sheath that surrounds the central pair of microtubules. Composed of a stalk that attaches to each doublet microtubule and a globular structure (spoke head) that projects toward the central pair of microtubules." [ISBN:0124325653, PMID:9450971]
> subset: goslim_pir
> xref: Wikipedia:Radial_spoke
> is_a: GO:0032991 ! protein-containing complex
> relationship: part_of GO:0005930 ! axoneme

**new stanza**

[Term]
id: GO:0001534
name: radial spoke
namespace: cellular_component
**def: "A protein structure involved in some types of beating motions of the cilium. The radial spoke is usually a T-shaped structure comprised of a short base that attaches to the A-microtubule of an axonemal outer microtubule doublet (MTD) of a cilium, an elongated stalk, a neck complex, and an orthogonal head structure that extends perpendicularly towards the inner sheath and the central pair (CP) microtubules. Groups of radial spokes (RSs) repeat along the MTD with regular spacing. In most organisms, each group of radial spokes is comprised of a triplet of spokes: RS1, RS2, and RS3. In some organisms (e.g. Chlamydomonas and Sarcophaga bullata), each group contains a doublet of radial spokes: RS1 and RS2, while RS3 is represented only as a stump (referred to as RS3S) attached to the A-microtubule but lacking the rest of the stalk structure and entirely lacking the head structure." [GOC:krc, ISBN:0124325653, PMID:22754630, PMID:25694453, PMID:34871179, PMID:9450971]**
subset: goslim_pir
xref: Wikipedia:Radial_spoke
is_a: GO:0032991 ! protein-containing complex
relationship: part_of GO:0005930 ! axoneme
**relationship: part_of GO:0097729 ! 9+2 motile cilium
property_value: term_tracker_item "https://github.com/geneontology/go-ontology/issues/25321" xsd:string**

----

**old stanza**

> [Term]
> id: GO:0001535
> name: radial spoke head
> namespace: cellular_component
[...]

PR #25350, opened 2023-05-09
Revised definition of radial spoke. #25321

raymond91125:


krchristie:
Regarding the term for 'radial spoke': 

> def: "Protein complex that links the outer microtubule doublet of the ciliary or flagellum axoneme with the sheath that surrounds the central pair of microtubules. Composed of a stalk that attaches to each doublet microtubule and a globular structure (spoke head) that projects toward the central pair of microtubules." [ISBN:0124325653, PMID:9450971]

I think we should 
* specify that radial spokes are part of 9+2 motile cilia within the definition
* add all the references I added to the definition as well as to the comment
* add this relationship: relationship: part_of GO:0097729 ! 9+2 motile cilium

krchristie:
I think it would be good to add the SubClass relationship
* part_of GO:0097729 ! 9+2 motile cilium

We've never created different subclasses of axonemes, so we don't have a term specifically for "9+2 axoneme". Therefore, to specify that the radial spoke is part of the axoneme specifically in 9+2 cilia, we should really add this additional part_of relationship.

This second comment is less important, but although I've looked at this definition over and over, I only noticed this with the addition of the "9+2" adjective, but this phrase: "the 9+2 ciliary or flagellum axoneme" is grammatically inconsistent since ciliary is the adjective form and flagellum is the noun. It would also sound better to use "a" instead of "the"

So, could we change this phrase to this: "a ciliary or flagellar axoneme"

raymond91125:
If there is no further request, I will merge. Thanks.

raymond91125:
I've added
9+2 motile cilium
relationship: has_part GO:0001534 ! radial spoke

9+2 non-motile cilium
relationship: has_part GO:0001534 ! radial spoke

raymond91125:
Good to merge? @krchristie Thanks.

ukemi:
Do all 9+2s have radial spokes?

raymond91125:
> Do all 9+2s have radial spokes?

That is my understanding but I don't think there is a definitive statement. It kind of makes sense, too, to have the radial spoke connecting the outer and inner MTs.

raymond91125:
Kinocilium is an example of non-motile 9+2 cilium in cochlear hair cells with radial spokes. https://www.frontiersin.org/articles/10.3389/fcell.2021.715037/full.

krchristie:
> > Do all 9+2s have radial spokes?
> 
> That is my understanding but I don't think there is a definitive statement. It kind of makes sense, too, to have the radial spoke connecting the outer and inner MTs.

My understanding is that all 9+2 motile cilia have radial spokes as radial spokes are essential for types of cilium movement found in 9+2 motile cilia. The radial spokes regulate dynein arm sliding along MTs to create bending in the axoneme and thus movement.

I don't currently remember any non-motile 9+2 cilia other than the kinocilia that @raymond91125 posted the link to the paper about. The paper says that the reason kinocilia are non-motile is because they lack inner dynein arms. I don't think there's any structural reason with respect to axoneme stability that radial spokes are required in 9+2 cilia as the central pair and the radial spokes are completely eliminated in 9+0 cilia (both motile and non-motile types). Thus, It might be safer to move this relationship:

9+2 non-motile cilium
relationship: has_part GO:0001534 ! radial spoke

down to the kinocililum term instead of on the parent term '9+2 non-motile cilium'

krchristie (approved):

PR #25402, opened 2023-05-18
Update go-upper.obo

pgaudet:
Hi @balhoff 
Alex needs to use this to be able to use the new SO term (not sure exactly where that is being used, so I cannot give you more explanation)

Thanks, Pascale


pgaudet:
That would be great!

PR #25690, opened 2023-07-05
Update new_pr.yml

hrshdhgd:
This PR:
- [x] Install the latest `ontobot-change-agent` in the workflow
- [x] Passes the prefix `GO` needed for new term request template processing (will be inactive for this repository).

### DIFF
--- a/.github/workflows/new_pr.yml
+++ b/.github/workflows/new_pr.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install ontobot-change-agent==0.3.11
+          pip install ontobot-change-agent
 
       - name: Run ochange
         id: ochange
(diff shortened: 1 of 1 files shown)


balhoff:
@hrshdhgd it seems like the QC action run for your fork does not read the `ODK_VERSION` variable we have set here. Do you know if that can be made to work?

https://github.com/geneontology/go-ontology/blob/3177889a3d4bc6bf6861e10f3f176b8749b1421a/.github/workflows/qc.yml#L22

hrshdhgd:
Aah! Let me fix this on this repo rather than opening a PR from my fork

hrshdhgd:
@balhoff , where is the `vars` variables defined? The GH workflows on my fork are running unnecessarily and sending me failure notifications. How do I stop these auto-runs? Any thoughts?

balhoff:
> @balhoff , where is the `vars` variables defined? The GH workflows on my fork are running unnecessarily and sending me failure notifications. How do I stop these auto-runs? Any thoughts?

Defined here: https://github.com/geneontology/go-ontology/settings/variables/actions

Is there a setting in your fork to turn off actions?

hrshdhgd:
Thanks! I think there is a setting to disable actions altogether but I'll need to run the ontobot one for debugging so it seems to be all or nothing. But I may be wrong. For now I've declared the variable in my repo too.

PR #25752, opened 2023-07-14
taxon-removed

ValWood:


ValWood:
@balhoff @pgaudet 
I merged this pull request. I don't think I should have. I was waiting for feedback and then got mixed up with this one:
https://github.com/geneontology/go-ontology/pull/25756
Le me know if I need to do anything

balhoff:
Do you want to restore it? I can revert it if so.

ValWood:
it might be ok. I only removed some constraints (intended).  I checked the file in and did a pull request (not sure if I needed to do that?) I couldn't figure out how to do the "regenerating?" part of the process. This is a heads up in case I caused a problem.

balhoff:
The regenerating will happen automatically once a day. Yes, always do a pull request for any sort of change.

ValWood:
OK it should be fine. I'm fixing the taxon constraint pre obsoletion now. I think that might be why I ran into issues before. I'll assume it's OK.

Issue #25698, opened 2023-07-06
NTR: [T cell exhaustion]


nana-amazs:
Please provide as much information as you can: 

* **Suggested term label:**
T cell exhaustion

* **Definition (free text)**
A state of T cell dysfunction due to chronic antigen stimulation in which T cells lose their effector functions and ability to kill cells 

* **Reference, in format PMID:#######**
•	PMID: 21739672
•	PMID: 31570879
•	PMID: [32117960](https://pubmed.ncbi.nlm.nih.gov/32117960)

* **Gene product name and ID to be annotated to this term***

NFAT ([Q8NET5](https://www.uniprot.org/uniprotkb/Q8NET5/entry)) 
(TOX) ([O94900](https://www.uniprot.org/uniprotkb/O94900/entry))
Eomes, ([O95936](https://www.uniprot.org/uniprotkb/O95936/entry))
T-bet, [Q9UL17](https://www.uniprot.org/uniprotkb/Q9UL17/entry)
BATF, [Q16520](https://www.uniprot.org/uniprotkb/Q16520/entry)
Blimp-1 [O75626](https://www.uniprot.org/uniprotkb/O75626/entry)
PD-1 [Q15116](https://www.uniprot.org/uniprotkb/Q15116/entry)
CTLA-4 [P16410](https://www.uniprot.org/uniprotkb/P16410/entry)
TIM-3 [P35625](https://www.uniprot.org/uniprotkb/P35625/entry)
LAG-3 [P18627](https://www.uniprot.org/uniprotkb/P18627/entry)
TIGIT [Q495A1](https://www.uniprot.org/uniprotkb/Q495A1/entry)

* **Parent term(s)**
GO:0042098 T cell proliferation
GO:0042110 T cell activation
GO:0046651 lymphocyte proliferation
GO:0002667 regulation of T cell anergy 
GO:0002668 negative regulation of T cell anergy
GO:0002710 negative regulation of T cell mediated immunity 
GO:0002698 negative regulation of immune effector process
GO:0050777 negative regulation of immune response

* **Children terms (if applicable)** Should any existing terms that should be moved underneath this new proposed term?
n/a

* **Synonyms (please specify, EXACT, BROAD, NARROW or RELATED)**
T cell Anergy (related)
T cell dysfunction (exact)
T cell Tolerance (related)

* **Cross-references**
KEGG

pgaudet:
[...]

PR #25760, opened 2023-07-15
Created exhausted T cell differentiation. #25698

raymond91125:


raymond91125:
@balhoff Jim, I'm stuck. I tried to resolve the conflicts but now my local repo is all screwed up. Could you merge for me? I will have to delete my local copy and start from scratch, I think.

balhoff:
@raymond91125 I think it might be easiest to start a new branch from `master`. You can probably even just copy and paste your new text from here into the OBO file, open, and re-save to get your term in the right order.

PR #26259, opened 2023-10-11
obsolete GO:0022889; Fixes #26052

### LINKED ISSUE #26052: GO:0022889 serine transmembrane transporter activity should explicitly use L-form
```yaml
id: GO:0022889
name: serine transmembrane transporter activity
namespace: molecular_function
def: "Enables the transfer of serine from one side of a membrane to the other. Serine is 2-amino-3-hydroxypropanoic acid." [GOC:mtg_transport, ISBN:0815340729, RHEA:35031]
xref: RHEA:35031
is_a: GO:0015175 ! neutral L-amino acid transmembrane transporter activity
intersection_of: GO:0022857 ! transmembrane transporter activity
intersection_of: has_primary_input CHEBI:35243 ! serine zwitterion
```

it was obvious to the person making this we were talking about the L-form, right? (spoiler: dangerous assumption, there are D-forms in the brain and in bacteria, there is a real possibility for mis-annotation here).

The is_a parentage makes it clear it's L-form.

And if we follow the link to RHEA it's clear it's L-form (L-serine zwitterion at physiological pH). https://www.rhea-db.org/rhea/35031

But the actual CHEBI ID we reference is weirdly specific on the protonation axis and weirdly general on the stereoisomer axis.

Ideally CHEBI would provide protonation neutral terms and we would invert this - be specific on stereoisomery and agnostic on protonation. However, they don't, so the correct course of action here is:

1. Rename this to be L-serine TTA (keep serine as a non-exact synonym)
2. change the text definition to explicitly say L-serine
3. change the logical definition to be L-serine zwitterion consistent with RHEA

edwong57:


cmungall:
Looks like this obsoletion was performed in a separate PR

- #26275

Issue #13926, opened 2017-07-17
regulates o part_of->regulates is in go_plus


ukemi:
We should not be inferring that if occurent1 regulates occurent2 and occurent2 is part_of occurent3, then occurent1 regulates occurent3. This is the case in go_plus and results in inaccurate inferred annotations generated by GO-CAM models, but these property chains are not in RO.

These also exist for the positive and negative properties.

balhoff:
@cmungall first item flushed out by the explanations view.

ukemi:
@vanaukenk  I think this will also address the issue with the Wnt model.

The model @balhoff and I looked at was this one:
http://noctua.berkeleybop.org/editor/graph/gomodel:586fc17a00001081

vanaukenk:
@balhoff and @ukemi  - thanks.  
@balhoff  - I didn't quite understand your comment above.  Has this property chain been fleshed out, or is it on the chopping block to be fleshed out?  The Wnt model still shows the inferences:
http://noctua.berkeleybop.org/editor/graph/gomodel:5966411600000538

balhoff:
@vanaukenk I just meant that this was the first problem explained by the explanations view. Still needs to be fixed!

dosumis:
Long standing ticket on flushing #12344.  Agree needs to be v. high priority. 

TL;DR:

Chain removed from RO.  
Before removing from GO we should instantiate the existing inferences as most are correct.  If we tag these in some way they could be reviewed and bad ones weeded out.

ukemi:
@dougli1sqrd will query for all of the inferred regulates relationships based on this property chain. The next step will be for an editor to go through them and determine which ones should be asserted. Then we will assert the correct ones and drop this property chain from GO-PLUS.

ukemi:
This project was discussed at the hackathon. We determined that if process/functionA regulated process/functionB and functionB is a part of processC, then process/functionA only regulates processC if process/functionA is causally_upstream_of and not part_of processC.

This means that the property chain regulates_o_part_of is not always valid. @balhoff determined that there are about 1700 relationships that will be lost when we remove this property chain. There are two ways forward: 1) Assert all of the currently inferred regulates relationships in the ontology and then remove the property chain. 2) Manually check the inferred relationships and add them by hand to the ontology. I prefer the latter.

[...]

PR #26278, opened 2023-10-13
obsoleted terms fixes #26165

### LINKED ISSUE #26165: Add references for  GO:0043963 perturbation of host adenylate cyclase-mediated signal transduction
GO:0043963 perturbation of host adenylate cyclase-mediated signal transduction
GO:0043965 symbiont-mediated suppression of host adenylate cyclase-mediated signal transduction

No annotations, no reference. 

Might correspond to PATHGO:0000174  regulates host cell adenylate cyclases/phosphodiesterases in another organism ??

pgaudet:


pgaudet:
hi @balhoff 

There is a massive diff because Protege 5.6.3 reformats a number of lines, for example:

> -relationship: positively_regulates GO:0009887 {http://purl.obolibrary.org/obo/IAO_0000233="https://github.com/geneontology/go-ontology/issues/13926"} ! animal organ morphogenesis
+relationship: positively_regulates GO:0009887 {IAO:0000233="https://github.com/geneontology/go-ontology/issues/13926"} ! animal organ morphogenesis

Is it OK ?

balhoff:
@pgaudet those `IAO:0000233` changes are expected due to the OWL API change I mentioned to you. So I think this is okay. However since there are so many changes we should probably have everyone move to the new version at the same time (more changes than I realized). Otherwise we may have this continually changing back and forth.

balhoff:
But there is also a problem in the CI check: `INVALID ELEMENT ERROR "http://www.geneontology.org/formats/oboInOwl#IAO:0000233" contains invalid characters`

What I think this is telling us is that we need to move to the new ROBOT (in the new ODK image) at the same time (or first).

PR #26392, opened 2023-11-08
Create design pattern for signaling receptor activity by activator.

balhoff:
This should end up grouping terms like [acetylcholine receptor activity](http://purl.obolibrary.org/obo/GO_0015464).


balhoff:
@pgaudet if you think there is a better term for the variable than "activator", e.g., maybe "input", I can change.

Issue #26426, opened 2023-11-14
Changes in upstream file locations breaks ontology build


kltm:
At least one mapping file has changed location and is breaking the build

Was
https://ftp.ebi.ac.uk/pub/databases/Rfam/CURRENT/rfam2go/rfam2go

Now (apparently)
https://ftp.ebi.ac.uk/pub/databases/Rfam/CURRENT/rfam2go

Tagging @balhoff

kltm:
Making https://github.com/geneontology/go-ontology/pull/26427

kltm:
Fixed

balhoff:
Thanks!

PR #26427, opened 2023-11-15
Update Makefile for new rfam2go location

kltm:
https://github.com/geneontology/go-ontology/issues/26426 Tagging @balhoff


kltm:
Gunna go for it, as all pipelines are down waiting on this.

Issue #26444, opened 2023-11-16
NTR: pH-gated channel activity terms


hattrill:
New terms for pH-gated ion channels

There is one pH-gated ion channel term: 
pH-gated chloride channel activity GO:0061797

it is under the term gated channel activity GO:0022836

Requesting terms for:
pH-gated sodium channel activity
and
pH-gated calcium channel activity

* **pH-gated calcium channel activity term**

A gated channel activity that enables the transmembrane transfer of a calcium ion by a channel that opens in response to a change in pH. 

is_a gated channel activity GO:0022836
is_a calcium channel activity GO:0005262

* **pH-gated sodium channel activity term**

A gated channel activity that enables the transmembrane transfer of a sodium ion by a channel that opens in response to a change in pH. 

is_a gated channel activity GO:0022836
is_a sodium channel activity GO:0005272

* **Gene product name and ID to be annotated to this term**
Recs1; FBgn0038209

Not sure if new parent term is needed or not.

pgaudet:
@hattrill can you please provide references for these activities?

hattrill:
Sorry, @pgaudet: PMID:34767445

raymond91125:
+[Term]
+id: GO:0160125
+name: pH-gated sodium channel activity
+namespace: molecular_function
+def: "A gated channel activity that enables the transmembrane transfer of a sodium ion by a channel that opens in response to a change in pH." [PMID:34767445]
+is_a: GO:0022836 ! gated channel activity
+relationship: has_primary_input CHEBI:29101 ! sodium(1+)
+property_value: term_tracker_item "" xsd:anyURI
+created_by: rynl
+creation_date: 2023-11-21T00:50:09Z
+
+[Term]
+id: GO:0160126
+name: pH-gated calcium channel activity
+namespace: molecular_function
+def: "A gated channel activity that enables the transmembrane transfer of a calcium ion by a channel that opens in response to a change in pH." [PMID:30930064, PMID:34767445]
+is_a: GO:0022836 ! gated channel activity
+relationship: has_primary_input CHEBI:29108 ! calcium(2+)
+property_value: term_tracker_item "https://github.com/geneontology/go-ontology/issues/26444" xsd:anyURI
+created_by: rynl
+creation_date: 2023-11-21T00:56:00Z

pgaudet:
@hattrill 

Looking at the paper again, it seems this is not 'gating', but regulation. The paper talk about 

> pH-dependent calcium and sodium channel

[...]

PR #26526, opened 2023-11-30
Added LDs. #26444

raymond91125:


pgaudet (changes requested):
HI @raymond91105 

Since the proton does not bind to the channel, it is not an activator. 

You need to use genera = pH-gated monoatomic ion channel; 
primary inputs are OK. 

Thanks, Pascale

raymond91125:
@pgaudet please review and approve changes if OK. thanks.

pgaudet (approved):

pgaudet:
Looks good!

PR #26527, opened 2023-11-30
Update only_in_taxon.tsv

ValWood:
The webpage says

"Make sure you do not add an extra line at the end of the file."

but  there was already a blank line at the end of the file. I left it without one, so if there is an issue this could be it.
@pgaudet


ValWood:
@pgaudet  this won't work anyway. I just remembered I need to add a tab for the ID, will fix

ValWood:
Unexpectedly it did work

Issue #26564, opened 2023-12-04
Unsynchronized changes between ontology and Makefile cause build (and pipeline) failures


kltm:
Changes to ontology https://github.com/geneontology/go-ontology/commit/a062b9e8675edb3c8e272dba61ad1a7421e8d4c2 are causing a miss with the Makefile target `prokaryote_subset`:

```
10:50:58  ROBOT_JAVA_ARGS=-Xmx48G robot --catalog catalog-v001.xml annotate -i build/subsets/prokaryote_subset.owl --annotation owl:versionInfo 2023-12-04 -o subsets/prokaryote_subset.owl convert -o subsets/prokaryote_subset.obo convert -o subsets/prokaryote_subset.json
10:50:58  org.semanticweb.owlapi.io.OWLOntologyInputSourceException: java.io.FileNotFoundException: build/subsets/prokaryote_subset.owl (No such file or directory)
```

kltm:
Tagging @balhoff and @pgaudet .
As part of changes like this, it might be good to have an SOP or crosscheck.

kltm:
https://github.com/geneontology/go-ontology/pull/26565

ValWood:
Is this still an issue?

balhoff:
I think it's taken care of.

PR #26565, opened 2023-12-04
Update makefile for goslim_prokaryote.

balhoff:
Fix ontology build failure. For #26564.


kltm:
https://github.com/geneontology/go-ontology/issues/26564

Issue #26618, opened 2023-12-12
NTR children of rRNA pseudouridine synthase activity (was: Add RHEA:54568 xref to GO:0120159 (rRNA pseudouridine synthase activity)) 


sjm41:
Add RHEA:54568 xref to GO:0120159 and update def to match:
a uridine in rRNA = a pseudouridine in rRNA

Should also consider adding the child terms corresponding to the specific EC/RHEA reactions that specify the rRNA and uridine position 

id: GO:0120159
name: rRNA pseudouridine synthase activity
namespace: molecular_function
def: "Catalysis of the reaction: an rRNA uridine = an rRNA pseudouridine. Conversion of uridine in an rRNA molecule to pseudouridine by rotation of the C1'-N-1 glycosidic bond of uridine in RNA to a C1'-C5." [PMID:28432181]
synonym: "16S rRNA pseudouridine(516) synthase" NARROW [EC:5.4.99.19]
synonym: "21S rRNA pseudouridine(2819) synthase" NARROW [EC:5.4.99.43]
synonym: "23S rRNA pseudouridine(1911/1915/1917) synthase" NARROW [EC:5.4.99.23]
synonym: "23S rRNA pseudouridine(2457) synthase" NARROW [EC:5.4.99.20]
synonym: "23S rRNA pseudouridine(2604) synthase" NARROW [EC:5.4.99.21]
synonym: "23S rRNA pseudouridine(2605) synthase" NARROW [EC:5.4.99.22]
synonym: "23S rRNA pseudouridine(746) synthase" NARROW [EC:5.4.99.29]
synonym: "23S rRNA pseudouridine(955/2504/2580) synthase" NARROW [EC:5.4.99.24]
xref: EC:5.4.99.19
xref: EC:5.4.99.20
xref: EC:5.4.99.21
xref: EC:5.4.99.22
xref: EC:5.4.99.23
xref: EC:5.4.99.24
xref: EC:5.4.99.29
xref: EC:5.4.99.43
is_a: GO:0009982 ! pseudouridine synthase activity
is_a: GO:0140098 ! catalytic activity, acting on RNA
property_value: term_tracker_item "https://github.com/geneontology/go-ontology/issues/15632" xsd:anyURI
created_by: krc
creation_date: 2018-05-11T21:24:19Z

sjm41:
Also add these:

RHEA:48348
a uridine in RNA = a pseudouridine in RNA
[...]

PR #26724, opened 2024-01-09
Removed EC xrefs and synonyms from parent

sjm41:


pgaudet:
for https://github.com/geneontology/go-ontology/issues/26618

Issue #26122, opened 2023-09-13
alpha-ketoglutarate (aka 2-oxoglutarate) dehydrogenase complex terms


sjm41:
Currently we have:

```
dihydrolipoyl dehydrogenase complex (GO:0045240)
|__cytosolic alpha-ketoglutarate dehydrogenase complex (GO:0045241)
|__mitochondrial alpha-ketoglutarate dehydrogenase complex (GO:0005947)
	|__mitochondrial oxoglutarate dehydrogenase complex (GO:0009353)
|__oxoglutarate dehydrogenase complex (GO:0045252)
        |__cytosolic oxoglutarate dehydrogenase complex (GO:0045248)
        |__mitochondrial oxoglutarate dehydrogenase complex (GO:0009353)
```

There are a few issues here:

1. The parent term (dihydrolipoyl dehydrogenase complex) is defined as "A protein complex that possesses alpha-ketoglutarate dehydrogenase activity.", so I'd suggest renaming the term to "alpha-ketoglutarate dehydrogenase complex" to match the definition. That said, I'm not convinced we need this grouping term at all. Obsolete?

2. There are zero annotations to and there is no reference for "cytosolic alpha-ketoglutarate dehydrogenase complex (GO:0045241)".  Obsolete?

3. 'mitochondrial alpha-ketoglutarate dehydrogenase complex (GO:0005947)' seems like a confusing and unnecessary grouping term with only one child term. So I'd obsolete GO:0005947 too.

4. There are 15 direct EXP annotations using GO:0005947 for human/rat/cerevisiae proteins.  All the human and rat proteins are components of the 'branched-chain alpha-keto acid dehydrogenase' complex (PMID: 3593587, PMID: 9582350, PMID: 10745006), so I think we need a new GO term specifically for that complex - see #26121.  The remaining 4 yeast annotations from PMID:25165143 (made by ComplexPortal) should be moved to the current child term mitochondrial oxoglutarate dehydrogenase complex (GO:0009353).

deustp01:
> 2\. There are zero annotations to and there is no reference for "cytosolic alpha-ketoglutarate dehydrogenase complex (GO:0045241)".  Obsolete?

[...]

PR #26827, opened 2024-01-18
Renamed and merged terms. #26122

raymond91125:


raymond91125:
@balhoff Jim, I got the error message of obsoleted term having a superclass. But I cannot figure out how that is the case. Could you help? Thanks.

FAIL Rule ../sparql/obsolete-reference-violation.sparql: 1 violation(s)
obsolete,reason
http://purl.obolibrary.org/obo/GO_0005947,Obsolete term should not have a superclass

PR #26889, opened 2024-01-26
Remove duplicate xref from obsolete term.

balhoff:
http://purl.obolibrary.org/obo/GO_0010356 and http://purl.obolibrary.org/obo/GO_0102551 both had an xref to RHEA:38003. But GO_0102551 is obsolete and replaced_by GO_0010356. There is a robot repair step that moves xrefs from obsolete terms to their replacements, which I think is what let this pass CI. But I'm not sure why it's failing when running in the pipeline.


kltm:
https://github.com/geneontology/go-ontology/issues/26890

kltm:
Pass.

Issue #26881, opened 2024-01-26
NTR: "branched-chain alpha-keto acid dehydrogenase complex" terms


sjm41:
Hi @raymond91105 

Related to #26122, please make the following two terms:

name: branched-chain alpha-keto acid dehydrogenase complex
namespace: cellular_component
def: "A multi-enzyme complex that catalyzes the oxidative decarboxylation of branched-chain alpha-keto acids derived from L-leucine, L-isoleucine, and L-valine to branched-chain acyl-CoAs. The complex comprises three main components: branched-chain alpha-ketoacid dehydrogenase (E1, consisting of alpha and beta subunits), dihydrolipoyl transacylase (E2), and dihydrolipoyl dehydrogenase (E3)." [PMID:10745006, PMID:10464218]
comment: Note that this term represents a location and not a function; the activity possessed by this complex is mentioned in the definition for the purpose of describing and distinguishing the complex. The catalytic activities of the individual components of this complex are represented by the molecular function terms '3-methyl-2-oxobutanoate dehydrogenase (2-methylpropanoyl-transferring) activity ; GO:0003863' (E1), 'dihydrolipoyllysine-residue (2-methylpropanoyl)transferase activity ; GO:0043754' (E2), and 'dihydrolipoyl dehydrogenase activity ; GO:0004148' (E3). The overall catalytic activity of the complex is represented by the molecular function term 'branched-chain alpha-keto acid dehydrogenase activity (GO:0047101)'.
is_a: GO:0045240 ! alpha-ketoacid dehydrogenase complex
intersection_of: GO:0032991 ! protein-containing complex
intersection_of: capable_of GO:0047101 ! branched-chain alpha-keto acid dehydrogenase activity

name: cytosolic branched-chain alpha-keto acid dehydrogenase complex
namespace: cellular_component
def: "A complex that catalyzes the oxidative decarboxylation of branched-chain alpha-keto acids derived from L-leucine, L-isoleucine, and L-valine to branched-chain acyl-CoAs.  It is composed of four polypeptides, E1alpha, E1beta, E2, and E3. [PMID: 10464218]
[...]

PR #26901, opened 2024-01-29
Obsoleted multiple terms. #26881

raymond91125:


raymond91125:
@balhoff Hi Jim, I'm stunned by this error:
org.semanticweb.owlapi.io.UnparsableOntologyException: Problem parsing file:/__w/go-ontology/go-ontology/src/ontology/go-edit-trimmed-imports.obo
Could not parse ontology.  Either a suitable parser could not be found, or parsing failed.  See parser logs below for explanation.
I don't know that I've changed this file.

balhoff:
@raymond91125 it's not your fault—a mistake got into the `master` branch via a direct push. I made a fix in a PR and merged it. So you should be able to update your branch from `master` and fix this. Also, we have modified the users who had admin permissions to try to prevent these issues. So if you have any permissions trouble in the near future, let us know.

PR #27313, opened 2024-03-15
Obsolete GO:0031668 cellular response to extracellular stimulus; Fixes #27020

### LINKED ISSUE #27020: Obsoletion request: 'GO:0031668 cellular response to extracellular stimulus'  and 'GO:0009991 - response to extracellular stimulus'
'GO:0031668 cellular response to extracellular stimulus' >> this term is too vague. 

Replace by [cellular response to stimulus](http://purl.obolibrary.org/obo/GO_0051716)

Thanks, Pascale

edwong57:
There were some instances where GO:0031668 was used in logical definitions. I changed them to either GO:000991 - response to extracellular stimulus or GO:0051716 - cellular response to stimulus if the definition/term name specified extracellular or not.


pgaudet:
Hi @edwong57 
Here the 'required tests' dont fail. The check that fails is about 'inference diffs', ie the report that shows all changes. In this case there were a lot of changes, so the report could not be generated; that's why it shows an error. This can be merged. 

I added this to the documentation: https://wiki.geneontology.org/Interpretation_of_Travis_checks_errors

edwong57:
Need to obsolete child term, GO:0009991 as well. Will do them both together.

Issue #27316, opened 2024-03-16
Full ontology build fails in pipeline due to "missing file" gocheck_do_not_manually_annotate.owl 


kltm:
I believe there has been recent changes around the "do not annotate" stuff recently. It looks like we're now failing on something like a rename or filename issue?

```
17:08:34  ROBOT_JAVA_ARGS=-Xmx48G robot --catalog catalog-v001.xml annotate -i build/subsets/gocheck_do_not_manually_annotate.owl --annotation owl:versionInfo 2024-03-15 -o subsets/gocheck_do_not_manually_annotate.owl convert -o subsets/gocheck_do_not_manually_annotate.obo convert -o subsets/gocheck_do_not_manually_annotate.json
17:08:35  org.semanticweb.owlapi.io.OWLOntologyInputSourceException: java.io.FileNotFoundException: build/17:08:34  ROBOT_JAVA_ARGS=-Xmx48G robot --catalog catalog-v001.xml annotate -i build/subsets/gocheck_do_not_manually_annotate.owl --annotation owl:versionInfo 2024-03-15 -o subsets/gocheck_do_not_manually_annotate.owl convert -o subsets/gocheck_do_not_manually_annotate.obo convert -o subsets/gocheck_do_not_manually_annotate.json
17:08:35  org.semanticweb.owlapi.io.OWLOntologyInputSourceException: java.io.FileNotFoundException: build/subsets/gocheck_do_not_manually_annotate.owl (No such file or directory)
17:08:35  Use the -vvv option to show the stack trace.
17:08:35  Use the --help option to see usage information.
17:08:35  
17:08:35  real	0m0.972s
17:08:35  user	0m2.343s
17:08:35  sys	0m0.224s
17:08:35  make: *** [Makefile:307: subsets/gocheck_do_not_manually_annotate.owl] Error 1(No such file or directory)
17:08:35  Use the -vvv option to show the stack trace.
17:08:35  Use the --help option to see usage information.
17:08:35  
17:08:35  real	0m0.972s
17:08:35  user	0m2.343s
17:08:35  sys	0m0.224s
17:08:35  make: *** [Makefile:307: subsets/gocheck_do_not_manually_annotate.owl] Error 1
```

kltm:
According to @pgaudet on slack: "The subset has been removed".

kltm:
[...]

PR #27319, opened 2024-03-16
Update Makefile to remove variable for gocheck_do_not_manually_annotate

kltm:
For #27316

Could @balhoff or @pgaudet add documentation for this to a README or SOP so we don't lose the pipeline to these changes in the future?


balhoff (approved):
Thanks!

kltm:
This is now waiting on update w/ #27336

PR #27414, opened 2024-03-25
Moved subsets from kwalify to linkml and tidied yaml file.

### LINKED ISSUE #27400: Review subsets YAML and align with best practices
While we have a draft of the YAML for all subsets, (https://github.com/geneontology/go-ontology/blob/master/src/ontology/subsets/subsets.schema.yaml), we need to review it and bring it in like with best practices.

cmungall:
Fixes #27400

Updates subsets metadata and added a schema

Added metadata fields for

- archived_url
- see_also

Previous versions used a kwalify schema which is no longer supported.
Schema is now in LinkML which allows for more flexibility.

- added text descriptions for all fields
- added URIs for common metadata elements

LinkML also allows for rules so we can say "if a subset is active
then it must have a description, tracker, and a context" (ideally we would
place this constraint on all but there are legacy subsets we don't
have this for.

@alexsign you can use either the linkml or the jsonschema. Can also generate java classes, pydantic, etc if helpful

TODO (possibly other PRs):

- [ ] add checks to github actions (@balhoff @kltm we are using latest-ish version of odk? in which case linkml is included)
- [x] discuss with @pgaudet about label vs title
- [ ] make an OBO PURL that resolves to this

Note that validation is not yet integrated into gh actions. For now,
do this:

    linkml-validate -s subsets.schema.yaml go_subsets_metadata.yaml

Or simply

   make validate

### DIFF
--- a/src/ontology/subsets/Makefile
+++ b/src/ontology/subsets/Makefile
@@ -1,0 +1,12 @@
+# TODO: integrate this Makefile into normal gh actions.
+# for now you must do a separate pip or poetry install
+# of LinkML
+
+RUN = poetry run
+SCHEMA = subsets.schema.yaml
+
+validate: go_subsets_metadata.yaml $(SCHEMA)
+	$(RUN) linkml-validate -s subsets.schema.yaml $<
+
+subsets.jsonschema.json: subsets.schema.yaml $(SCHEMA)
+	$(RUN) gen-json-schema $< > $@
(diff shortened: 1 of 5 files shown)


kltm:
@cmungall from other thread:
`${{ vars.ODK_VERSION }}`
https://github.com/geneontology/go-ontology/settings/variables/actions
`v1.4`

pgaudet:
Can this be merged?

Issue #27399, opened 2024-03-22
Add GHA QC for subsets YAML schema


kltm:
We now have a YAML for all subsets, (https://github.com/geneontology/go-ontology/blob/master/src/ontology/subsets/subsets.schema.yaml), but we need to add QC checking for this.

From @cmungall , we can use linkml for this, and port it in like 
`schemauto import-kwalify my/schema/personinfo.kwalify.yaml`

kltm:
@pgaudet GHA for metadata is a requirement for the software side, so adding to project.

kltm:
@pgaudet I'll need some additional info on the PR (https://github.com/geneontology/go-ontology/pull/28196) to bring the metadata into line with the schema.

pgaudet:
hi @kltm 

> In go_subsets_metadata.yaml, goslim_virus has two roles; the schema requires a single role

OK, I can we have a subset have two roles? can I duplicate the entry? I would rather not create a copy of the subset itself. 
Or is this something we can change in the metadata spaces? I couldn't see that it was a single-value field.

> In go_subsets_metadata.yaml, goslim_goa needs the additional metadata (like contact, role) to conform with the schema, even though it is obsolete; this can be cribbed from the other obsolete subsets?

fixed. However I noticed that the subset is archived on the cvs repository, which is offline: http://cvsweb.geneontology.org/cgi-bin/cvsweb.cgi/go/GO_slims/archived_GO_slims/goslim_goa.2002?rev=1.4;content-type=text%2Fplain

Do we have a copy anywhere? Should we write that it's unavailable?

kltm:
@pgaudet 

* Re: roles <br/> we can have the metadata structured however we want, _but_, I'm not sure how it's getting used in the ontology build and whether it's legal; if it is indeed legal, we just switch it to the multivalued field
* Re: subset metadata <br/> as it's archived, I'm not too worried about it not being available, as that's what's going on anyways now; it would at least give somebody a heads-up on where to look if it came to that

kltm:
Oddly, the PR check (https://github.com/geneontology/go-ontology/actions/runs/9865845910/job/27243789727?pr=28196) failed with:
```
Run actions/checkout@v2
/usr/bin/docker exec  beb84bab745e276dd443df91ee52820308752c1778e2d5253758dbbe67bad09d sh -c "cat /etc/*release | grep ^ID"
[...]

PR #28196, opened 2024-06-18
add actual gha check

kltm:
for #27399


kltm:
Noting that this can be merged once the tests pass.

kltm:
Talking to @pgaudet , we are not using this internally, so it should be safe to make a schema change to allow a multi-valued role

kltm:
Merging, to prevent drift and the fail is issue https://github.com/geneontology/go-ontology/issues/28454

Issue #28336, opened 2024-06-27
Textual definition update: GO:0052811 GO:0000285 GO:0035005 GO:0016308 GO:0046934 GO:0016309 GO:0016303 GO:0004430 GO:0052810


rozaru:
For most of the terms bellow there is a typo: 2H+ should have been H+ and a "a" is missing 

id: GO:0052811
name: 1-phosphatidylinositol-3-phosphate 4-kinase activity

New definition:
Catalysis of the reaction: a 1-phosphatidyl-1D-myo-inositol 3-phosphate + ATP = a 1-phosphatidyl-1D-myo-inositol 3,4-bisphosphate + ADP + H+.

---
id: GO:0000285
name: 1-phosphatidylinositol-3-phosphate 5-kinase activity

New definition:
Catalysis of the reaction: a 1-phosphatidyl-1D-myo-inositol 3-phosphate + ATP = a 1-phosphatidyl-1D-myo-inositol 3,5-bisphosphate + ADP + H+.

---
id: GO:0035005
name: 1-phosphatidylinositol-4-phosphate 3-kinase activity

New definition:
Catalysis of the reaction: a 1-phosphatidyl-1D-myo-inositol 4-phosphate + ATP = a 1-phosphatidyl-1D-myo-inositol 3,4-bisphosphate + ADP + H+.

---
id: GO:0016308
name: 1-phosphatidylinositol-4-phosphate 5-kinase activity

New definition:
Catalysis of the reaction: a 1-phosphatidyl-1D-myo-inositol 4-phosphate + ATP = a 1-phosphatidyl-1D-myo-inositol 4,5-bisphosphate + ADP + H+.

---
id: GO:0046934
name: 1-phosphatidylinositol-4,5-bisphosphate 3-kinase activity

New definition:
Catalysis of the reaction: a 1-phosphatidyl-1D-myo-inositol 4,5-bisphosphate + ATP = a 1-phosphatidyl-1D-myo-inositol 3,4,5-trisphosphate + ADP + H+.

---
id: GO:0016309
name: 1-phosphatidylinositol-5-phosphate 4-kinase activity

New definition:
Catalysis of the reaction: a 1-phosphatidyl-1D-myo-inositol 5-phosphate + ATP = a 1-phosphatidyl-1D-myo-inositol 4,5-bisphosphate + ADP + H+

additional information:
This def for the reaction from EC 2.7.1.149 differs from how the other one were written.
[...]

PR #28356, opened 2024-06-28
Revised definitions; #28336

raymond91125:


raymond91125:
@balhoff How do I fix this problem "INVALID ONTOLOGY FILE ERROR Could not load a valid ontology from file: enhanced.owl". It is there, src/ontology/enhanced.owl. Thanks.

Issue #28454, opened 2024-07-10
In some cases, ontology build fails due to node js issues related to checkout workflow


kltm:
For example, the PR check (https://github.com/geneontology/go-ontology/actions/runs/9865845910/job/27243789727?pr=28196) failed with:
```
Run actions/checkout@v2
/usr/bin/docker exec  beb84bab745e276dd443df91ee52820308752c1778e2d5253758dbbe67bad09d sh -c "cat /etc/*release | grep ^ID"
/__e/node20/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)

Possibly: https://github.com/actions/checkout/issues/1590
```

This has continued, even though @balhoff tried v4 (https://github.com/geneontology/go-ontology/pull/28451)

Tagging @balhoff @pgaudet

kltm:
https://github.com/geneontology/go-ontology/tree/issue-28454-checkout-issue-w-node-js

kltm:
Trying things from https://github.com/actions/checkout/issues/1590

kltm:
@balhoff I have a possible solution here.
From the external thread, it seems that the "unsafe" trick does not work for v4, so I tried with v3. If you can live with this, we can merge.
I would also note, that this is believed to be a temporary fix, and there will be more issues down the road, as we track a moving image. We'll need to weight the trade-offs with that at some point, but for now I'd be happy with passing tests again.

balhoff:
@kltm possible fix in #28465 (different action used for posting PR comments).

kltm:
From @balhoff 's fixes, I think this is clear now.

PR #28455, opened 2024-07-10
try and fix checkout on github workflows

kltm:
for #28454


balhoff (commented):
I think we no longer need to make these changes, since I'm now using a different, improved, action for posting comments.

PR #28472, opened 2024-07-12
Update Makefile

### LINKED ISSUE #28471: Add gocheck_obsoletion_candidate to published subsets in GO makefile
Add gocheck_obsoletion_candidate to published subsets in GO makefile

pgaudet:
fixes #28471


balhoff (commented):
@pgaudet has the subset been added to the ontology already? (at least declared) I'm worried that the pipeline will break if this is added before it exists in the ontology.

pgaudet:
yes - but we can wait for the snapshot to go through, since we've been having so many problems.

Issue #21347, opened 2021-04-26
implement check for xrefs to obsolete Rhea or EC terms


balhoff:

balhoff:
See: https://ftp.expasy.org/databases/rhea/tsv/rhea-obsoletes.tsv

pgaudet:
For EC, this is the file that contains the valid ECs and the history: 
https://ftp.expasy.org/databases/enzyme/enzyme.dat

note that this updated at every UniProt release, so we only need to run it every 2 months. 

@amorgat @alexsign Could we have 'early' versions of those files, to make corrections before the release? 

Thanks, Pascale

balhoff:
Done for Rhea in #22240, but only reporting, not yet failing the build.

balhoff:
In order to complete this for Rhea, we should cache the Rhea file like other imports, and create separate PRs for migrating to new Rhea builds. Then we can have this check fail the build safely.

balhoff:
See also #20301.

balhoff:
Also consider MetaCYC.

pgaudet:
and KEGG ?

pgaudet:
Obsolete EC xrefs that need fixing (see https://github.com/geneontology/go-ontology/pull/27017):

- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0002950> has an xref to an obsolete EC: "EC:2.7.8.n3"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0004121> has an xref to an obsolete EC: "EC:4.4.1.8"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0004584> has an xref to an obsolete EC: "EC:2.4.1.130"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0050136> has an xref to an obsolete EC: "EC:1.6.5.11"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0008668> has an xref to an obsolete EC: "EC:2.7.7.58"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0008769> has an xref to an obsolete EC: "EC:3.4.13.3"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0008940> has an xref to an obsolete EC: "EC:1.7.99.4"
- [ ] ERROR: <http://purl.obolibrary.org/obo/GO_0016720> has an xref to an obsolete EC: "EC:1.14.99.33"
[...]

PR #28491, opened 2024-07-15
test obsolete RHEA for #21347

pgaudet:


balhoff:
This PR now fails as expected; closing.

PR #28492, opened 2024-07-15
test obsolete EC for #21347

pgaudet:


pgaudet:
Good - test is failing correctly.

PR #28533, opened 2024-07-18
Vw18 07 24

ValWood:


ValWood:
@pgaudet  I don't know why this opened a ticket on the ontology tracker do you? It doesn't usually?

PR #28699, opened 2024-08-14
Refresh ChEBI import.

balhoff:


balhoff:
Need to review these classification changes.

pgaudet:
Not great, but these are the normal ChEBI issues. Some of the hierrchy is improved (for eg, removal of '[phosphorus metabolic process](http://purl.obolibrary.org/obo/GO_0006793)' as a parent of many terms).

PR #29175, opened 2024-11-07
Quick and dirty test of chemessence import.

balhoff:


balhoff:
No unsatisfiable classes; diff is too big to post as a comment, and has a lot of useless stuff. I will try making a relation-graph-diff version.

Issue #29150, opened 2024-11-01
Obsoletion request: GO:0030374 nuclear receptor coactivator activity and replace by GO:0003714 transcription corepressor activity


pgaudet:
Dear all, 

The proposal has been made to obsolete 
GO:0030374 nuclear receptor coactivator activity and GO:0140536 nuclear receptor corepressor activity and replace by their respective parents, [GO:0003713 transcription coactivator activity] and [GO:0003714 transcription corepressor activity](https://amigo.geneontology.org/amigo/term/GO:0003714#display-lineage-tab). 

The reason for obsoletion is that these activities are not necessarily specific for nuclear receptors subclass of dbTFs, so that distinction leads to misannotations. 

All impacted groups annotations will be automatically re-annotated, see annotation ticket: https://github.com/geneontology/go-annotation/issues/5502

InterPro and UniRule/HAMAP annotations need to be moved up to the parent. 

Thanks, Pascale

@colinlog

pgaudet:
Removed part of the comment of [nuclear receptor activity](http://purl.obolibrary.org/obo/GO_0004879), as pointed out by @slaulederkind here: https://github.com/geneontology/go-annotation/issues/5502#issuecomment-2465230146

> When the nuclear receptor functions by binding to other transcription factors or transcription factor complexes, consider instead annotating to 'GO:0030374 ; nuclear receptor transcription coactivator activity' or GO:0140536 ; nuclear corepressor activity.

pgaudet:
Fixed textual definitions and logical definitions for nuclear receptors: 

Textual definitions: "A nuclear receptor activity regulated by **xx binding** and modulating the transcription of specific gene sets transcribed by RNA polymerase II.

Logical definitions: 
+intersection_of: GO:0004879 ! nuclear receptor activity
+intersection_of: RO:0012000 CHEBI:35341 ! has small molecule regulator (small molecule)

PR #29180, opened 2024-11-11
fixed comment, definitions, LDs, for #29150

pgaudet:


pgaudet:
New attempt: https://github.com/geneontology/go-ontology/pull/29185

Issue #28506, opened 2024-07-16
Term definitions and scope of GO:0052695 cellular glucuronidation and GO:0019390 glucuronoside biosynthetic process


rozaru:
GO terms:

> **GO:0052695	cellular glucuronidation**
> The modification of an organic chemical by the conjugation of glucuronic acid. The substances resulting from glucuronidation are known as glucuronosides (or glucuronides) and are often much more water-soluble than the non-glucuronic acid-containing precursor. 
> 
> **GO:0036234	deglucuronidation**
> The removal of glucuronic acid from a conjugated substrate.

 and 

> **GO:0019390	name: glucuronoside biosynthetic process**
> The chemical reactions and pathways resulting in the formation of glucuronosides, compound composed of a hydroxy compound linked to a glucuronate residue.
> **GO:0019391	name: glucuronoside catabolic process**
> The chemical reactions and pathways resulting in the breakdown of glucuronosides, compound composed of a hydroxy compound linked to a glucuronate residue.

Based on their definition and how proteins involved in these processes are annotated I am not sure in which context they should be used (or what is the difference between them). At the moment, I have the impression that they mean more or less the same.

![image](https://github.com/user-attachments/assets/322ded4d-1dde-4e2e-b551-be811bbd368b)

Glucuronides (or glucuronosides) are produced by linking glucuronic acid (GlcA) (provided by UDP-GlcA) to a molecule via a glycosidic bond.
The molecule can be:
- endogenous substances (e.g., bilirubin)
- drugs
- xenobiotics (e.g., environmental toxins)
- also the same reaction is part of the synthesis of:
- tetrasaccharide linkage used for glycosaminoglycan biosynthesis (Ser/Thr- Xyl - Gal - Gal- GlcA)
- part of the repeat disaccharide unit of Heparin/Heparan sulfate (HS) and chondroitin sulfate (CS)/dermatan sulfate (DS)

Examples of annotation:

PMID: 22294686 was used to create the GO term deglucuronidation. In the paper they characterised ABHD10. For the annotation based on this paper GO glucuronoside catabolic process instead/ of GO cellular glucuronidation was used.

GO:0019391	glucuronoside catabolic process	 IDA 	PMID: 22294686
GO:0052695	cellular glucuronidation			TAS		Reactome

For mouse Gusb:
GO:0030200	heparan sulfate proteoglycan catabolic process	(IMP)
GO:0019391	glucuronoside catabolic process		ECO:0000314	(IDA)
[...]

PR #29379, opened 2024-12-04
Reinstated glucuronoside catabolic process; #28506

raymond91125:


raymond91125:
Not needed after all.

PR #29566, opened 2025-01-21
Use ROBOT to create JSON ontology serializations

balhoff:
Reverting reversion in #28960.

@kltm @sierra-moxon I'm trying this again, after Sierra's fixes in ontobio (https://github.com/biolink/ontobio/issues/697).


kltm:
@balhoff If you're around and this fails, since it will be pretty easy to revert if there are problems, I think this is probably okay to go ahead with.

PR #29874, opened 2025-03-07
Obsoleted 5 terms, fixes #29438

raymond91125:


raymond91125:
@balhoff Hi Jim, I don't understand the error message:
http://purl.obolibrary.org/obo/GO_1990224,Obsolete term should not have a superclass
It doesn't have a superclass that I can see. Thanks.

balhoff:
@raymond91125 sorry about this confusing message; in this case the term may either have a taxon constraint or Rhea participants axioms. Do either of these apply?

raymond91125:
Ah, my bad, forgot to commit go-catalytic-activities-participants.owl. Thanks. @balhoff

PR #30358, opened 2025-06-02
part 29770-obs-action7

ValWood:


ValWood:
[protein C-linked glycosylation via 2'-alpha-mannosyl-L-tryptophan](http://purl.obolibrary.org/obo/GO_0018406) [label](http://www.w3.org/2000/01/rdf-schema#label) "obsolete protein C-linked glycosylation via 2'-alpha-mannosyl-L-tryptophan"

is still in my local copy

PR #30901, opened 2025-10-08
Add id and shorthand to oboInOwl file. Save RDF/XML with current OWL API

balhoff:
For https://github.com/owlcs/owlapi/issues/1169.


allenbaron:
Does `oboInOwl:date` also need to be added here?

It seems like the OWLAPI also uses oboInOwl:date which does not exist in this file. I usually see this appear in OWL files converted from OBO (when the OBO file includes a date field as described in the [The OBO Flat File Format Guide, version 1.4](https://owlcollab.github.io/oboformat/doc/GO.format.obo-1_4.html)) using ROBOT.

balhoff:
@allenbaron I looked at `date`, and I see this already in the oboInOwl file:

https://github.com/geneontology/go-ontology/blob/59141af60f7b0bdb0c0720a32f8376985377b57b/contrib/oboInOwl.obo#L80-L84

It's a bit similar to how xref is defined:

https://github.com/geneontology/go-ontology/blob/59141af60f7b0bdb0c0720a32f8376985377b57b/contrib/oboInOwl.obo#L86-L90

When OWL API writes out an `xref` in OWL, it uses `oboInOwl:hasDbXref`. I'm wondering if it's a bug that OWL API doesn't write `oboInOwl:hasDate` in OWL. One issue is that OWL API will write any unknown tags into the `oboInOwl` namespace (I think!). Chasing corner cases in OBO format is kind of a rabbit hole.

cmungall (approved):

PR #31317, opened 2026-01-09
Add check for Rhea parent-child relationships

balhoff:
For #30646.


pgaudet:
@balhoff 

Can we make the error "GO:XXXXX has a Rhea xref, "RHEA:10740", with child reaction, "RHEA:21356", that is not a narrow xref or is an xref on a term that is not a subclass" not fail the checks ? 

We should fix those but there are a too many. 

Instead: I thought we had said that if a RHEA is already in GO, we would not try to load it. 

What do you think ?

balhoff:
Note—the CI build needs to be cleaned up on this PR (redundancy).

Issue #30720, opened 2025-08-19
NTR: guanylylpyridinol adenylase activity


magrane:
Please provide as much information as you can: 

* **Suggested term label:**
guanylylpyridinol adenylase activity

* **Definition (free text)**
Catalysis of the reaction: guanylylpyridinol + ATP + H(+) = guanylylpyridinol-AMP + diphosphate

* **Reference, in format PMID:#######**
PMID:25882909

* **Gene product name and ID to be annotated to this term**
hcgE (UniProt D9PY12)

* **Parent term(s)**

* **Children terms (if applicable)** Should any existing terms that should be moved underneath this new proposed term?

* **Synonyms (please specify, EXACT, BROAD, NARROW or RELATED)**

* **Cross-references**
Rhea 85487
EC 2.7.7.110
KEGG R13520

* **Any other information**

raymond91125:
EC is present at ExplorEnz but not at ExPASy, cannot be added yet.

raymond91125:
+[Term]
+id: GO:0160301
+name: guanylylpyridinol adenylase activity
+namespace: molecular_function
+def: "Catalysis of the reaction: guanylylpyridinol + ATP + H+ = guanylylpyridinol-AMP + diphosphate." [PMID:25882909, RHEA:85487]
+xref: KEGG_REACTION:R13520
+xref: RHEA:85487 {source="skos:exactMatch"}
+is_a: GO:0016779 ! nucleotidyltransferase activity
+property_value: term_tracker_item "https://github.com/geneontology/go-ontology/issues/30720" xsd:anyURI
+created_by: rynl
+creation_date: 2025-08-21T22:49:37Z
+

magrane:
EC 2.7.7.110 now available on expasy: https://enzyme.expasy.org/EC/2.7.7.110

Issue #30718, opened 2025-08-19
NTR: pyridinol guanylyltransferase activity


magrane:
Please provide as much information as you can: 

* **Suggested term label:**
pyridinol guanylyltransferase

* **Definition (free text)**
Catalysis of the reaction: 6-carboxymethyl-3,5-dimethyl-4-hydroxypyridin-2-ol + GTP + H(+) = guanylylpyridinol + diphosphate

* **Reference, in format PMID:#######**
PMID: 24249552

* **Gene product name and ID to be annotated to this term**
hcgB (UniProt Q57912)

* **Parent term(s)**
GO:0070568 (guanylyltransferase activity)

* **Children terms (if applicable)** Should any existing terms that should be moved underneath this new proposed term?

* **Synonyms (please specify, EXACT, BROAD, NARROW or RELATED)**

* **Cross-references**
Rhea 85479
EC 2.7.7.109
KEGG R13518

* **Any other information**

raymond91125:
+[Term]
+id: GO:0160298
+name: pyridinol guanylyltransferase activity
+namespace: molecular_function
+def: "Catalysis of the reaction: 6-carboxymethyl-3,5-dimethyl-4-hydroxypyridin-2-ol + GTP + H+ = guanylylpyridinol + diphosphate." [PMID:24249552, RHEA:85479]
+xref: KEGG_REACTION:R13518
+xref: RHEA:85479 {source="skos:exactMatch"}
+is_a: GO:0070568 ! guanylyltransferase activity
+property_value: term_tracker_item "https://github.com/geneontology/go-ontology/issues/30718" xsd:anyURI
+created_by: rynl
+creation_date: 2025-08-21T00:46:39Z

raymond91125:
Unpublished EC is not allowed as an xref. @magrane Is it known when EC:2.7.7.109 will be published?

magrane:
EC 2.7.7.109 is now published so could be added as an xref: https://enzyme.expasy.org/EC/2.7.7.109

PR #31450, opened 2026-01-29
Added EC xrefs, #30718, #30720

raymond91125:


raymond91125:
@balhoff Hi Jim, how often do we update EC terms? Is there a way I can trigger the process myself? Thanks.

balhoff:
> @balhoff Hi Jim, how often do we update EC terms? Is there a way I can trigger the process myself? Thanks.

It's this action: https://github.com/geneontology/go-ontology/actions/workflows/refresh_obsolete_ec_file.yml

It runs every Monday morning. You can also trigger it from that webpage.

raymond91125:
@balhoff I think I still have not updated obsolete_ec.txt to include EC:2.7.7.110 as valid. It is live https://enzyme.expasy.org/EC/2.7.7.110 and in enzyme.rdf. Do I need to do something else? Thanks.

balhoff:
@raymond91105 this would need to be merged: https://github.com/geneontology/go-ontology/pull/31467

But let me try to figure out why the checks didn't run.

balhoff:
Oh, there was a GH actions outage today.

raymond91125:
https://github.com/geneontology/go-ontology/issues/31488

PR #31761, opened 2026-03-25
Enhancing agent instructions for checking external usage when obsoleting

cmungall:


balhoff (commented):
Should we say that an entry only matters if the context is `UsageContext.RELATIONSHIP_OBJECT`?

PR #32379, opened 2026-07-28
TEST: Expand mitochondrial fission definition with mechanistic detail

cmungall:
`GO:0000266 mitochondrial fission` currently has a purely outcome-based definition:

> The division of a mitochondrion within a cell to form two or more separate mitochondrial compartments.

That describes the result but says nothing about how the division happens, which makes it harder to distinguish from `GO:0008053 mitochondrial fusion` at the definition level — both are membrane remodelling events on the same organelle.

This adds a clause covering the machinery and the membrane order of operations, keeping the existing `PMID:11038192` xref.

Single-line change to one `def:` tag; no axiom, synonym, or hierarchy changes.


cmungall:
Ignore this: this is a test for the review agent

PR #32382, opened 2026-07-28
Add localisation and cofactor detail to fatty acid beta-oxidation definition

cmungall:
`GO:0006635 fatty acid beta-oxidation` currently describes the chain-shortening chemistry in good detail — CoA addition, successive two-carbon removals, the acetyl-CoA/propionyl-CoA endpoint — but says nothing about **where** the pathway operates or **what drives** each round of the cycle.

Both come up when curators are deciding between this term and its siblings, so it seems worth having in the definition rather than implicit.

Adds a closing sentence covering subcellular localisation and the cofactor dependency. Existing `GOC:mah` and `ISBN:0198506732` xrefs retained; no axiom, synonym, or hierarchy changes.


cmungall:
Closing: this was a deliberate test of the review agent, not a real curation request. The added sentence was wrong on purpose, and the review correctly identified both errors as CRITICAL.

Recording the result, since the point was to find out whether the reviewer catches semantically wrong biology that passes every CI check:

- Both planted errors caught. Wrong compartment (cytosol vs mitochondria/peroxisomes) and inverted redox chemistry (NADPH-dependent reductases vs FAD/NAD+ dehydrogenases).
- Every citation verified accurate: GO:0033539, GO:0033540, GO:0016507, GO:0036125, GO:0008670, GO:0033543. No hallucinated ids, no misdescribed axioms.
- It steelmanned the error before rejecting it, identifying 2,4-dienoyl-CoA reductase (GO:0008670) as the one genuinely NADPH-dependent reductase in the pathway and explaining why that does not rescue the claim.
- It found something the plant's author did not anticipate: retaining the existing GOC:mah / ISBN xrefs means the new assertions are attributed to sources that state the opposite. That was intended as keeping the diff clean; it is a provenance violation.
- It attacked the PR's stated premise, not just the text: the mitochondrial/peroxisomal distinction is already carried by NARROW synonyms on the children, and GO:0006635 is not eukaryote-restricted, so no single-compartment statement can be correct at that level.

No changes to keep. Closing.

PR #32483, opened 2026-08-19
Add noctua-allowed ChEBI subset

balhoff:
Generate the Noctua allow-list as the union of Rhea pH 7.3 terms and the GO ChEBI allow-list. Include the union in the GO-LEGO ChEBI extraction, annotate each selected class, and merge the same subset annotations into go-amigo after its full ChEBI import closure.

See [geneontology/noctua#1100](https://github.com/geneontology/noctua/issues/1100)


balhoff:
@ai4c-reviewer please check my update commit.

PR #32541, opened 2026-08-26
Rename GO:0043039 to tRNA charging, part_of protein biosynthetic process, #15375

raymond91125:
[Written by Claude]

Step 1 of the recommendation posted on #15375 ([comment](https://github.com/geneontology/go-ontology/issues/15375#issuecomment-5431742836)). Spine edits only: **no obsoletions, no annotation migration, no new terms.** Steps 2 and 3 of that recommendation are deliberately not here — they need curator sign-off.

## Changes, all on GO:0043039

| | |
|---|---|
| `name` | `tRNA aminoacylation` → **`tRNA charging`** |
| `synonym` | `"tRNA charging" EXACT` → `"tRNA aminoacylation" EXACT` (label/synonym swap) |
| `synonym` | **added** `"aminoacyl tRNA synthesis" EXACT` |
| `relationship` | **added** `part_of GO:0160307 protein biosynthetic process` |
| `is_a` | `GO:0006399 tRNA metabolic process` — **retained** |

The rename and the added synonym implement @pgaudet's [annotation call decision](https://github.com/geneontology/go-ontology/issues/15375#issuecomment-3347997731): *"part_of 'tRNA aminoacylation' >> rename 'tRNA charging'? add synonyms: aminoacyl tRNA synthesis, amino acid activation"*. `"amino acid activation"` was already present as a RELATED synonym and is untouched.

Also refreshed four stale `! tRNA aminoacylation` trailing label comments on terms that reference GO:0043039. Comment-only, no semantic change.

## The one substantive departure: keeping `is_a GO:0006399`

The tracked plan for this ticket ([item 4](https://github.com/geneontology/go-ontology/issues/15375#issuecomment-4173125702)) calls for **removing** this, per @ValWood: *"Remove: is_a GO:0006399 tRNA metabolic process (currently asserted — tRNA is not metabolized)"*. This PR keeps it. Three reasons:

**1. The term satisfies GO:0006399's own definition.** `tRNA metabolic process` is a defined class:

```obo
intersection_of: GO:0008152 ! metabolic process
intersection_of: has_primary_input_or_output SO:0000253 ! tRNA
```

Charging is `tRNA + amino acid + ATP -> aminoacyl-tRNA`. The tRNA is the primary input; an aminoacyl-tRNA — still a tRNA — is the primary output. Removing the assertion would not stop the term being a tRNA metabolic process. It would just stop it saying so.

[...]

Issue #31932, opened 2026-04-21
Review/obsolete all BP terms for specific/variant metabolic pathways


sjm41:
In other separate tickets, we have already obsoleted many BP terms describing specific variants of metabolic pathways, just keeping the more general parent term describing the overall purpose of the process. Such variant pathways (that differ in the start/end product, or the specific enzyme(s) used) are better captured as GO-CAMs if needed, and the alternative approach (i.e. creating new GO terms for every variant pathway) isn't scalable/useful. The majority of terms for specific variant pathways surveyed thus far had never been used in annotation anyway.

I attempted to systematically find all such BP terms so that we could review and potentially obsolete all such terms systematically.

I combined the results of 2 different approaches:

1. Searh for BP terms whose label contained one of these strings:
process to
process via
process, by
salvage from
oxidation to
decarboxylation via
process using
process through
pathway through
process, using
oxidation using
oxidation,
oxidation via
process,
fermentation to
fermentation via

2. With Claude's help, I searched for BP terms that had an (implied) exactMatch xref to a MetaCyc 'variant' pathway, identified as MetaCyc pathways with names that end with a roman numeral (not 100% accurate, but pretty good).

Resulting list to review is here:
https://docs.google.com/spreadsheets/d/1HngvKWzFEFTyYrPB48rcku7sc-aPJvyYNpl-Ydfr6dM/edit?pli=1&gid=0#gid=0

Summary:
- 177 candidate terms to review
- 105 of those have 0 EXP annotations
- a further 39 terms have 5 or fewer EXP annotations
- 33 terms have 6 or greater EXP annotations, suggesting they should be more carefully reviewed (several of these have already been looked at and should be kept)

pgaudet:
Review for pathway variants terms with 3 or less annotations: 
https://docs.google.com/spreadsheets/d/1AXlETj5r1WTq1cMJk1csH9lv4EBWhI5NmU6bXaj5ZFY/edit?gid=0#gid=0

sjm41:
Following merges above, still need to add the following MetaCyc IDs as narrowMatch xrefs to some terms:

- MetaCyc:DARABCAT-PWY	GO:0019571 D-arabinose catabolic process
- MetaCyc:P185-PWY	GO:0019649 formaldehyde assimilation
- MetaCyc:PWY-1187	GO:0019761 glucosinolate biosynthetic process
- MetaCyc:PWY-1861	GO:0019649 formaldehyde assimilation
[...]

Issue #15375, opened 2018-03-10
OBSOLETE: GO:0043038 amino acid activation GO:0043039    tRNA aminoacylation GO:0043041    amino acid activation for nonribosomal peptide biosynthetic process


Antonialock:
should 

GO:0006426 glycyl-tRNA aminoacylation &
GO:0006544 glycine metabolic process

be related?

hdrabkin:
No I don't think so.  We don't do this for the other tRNA aminoacylations. It's the old question of where we draw the line for substrates. Aminoacyl-trna aminoacylation are also not related to ATP metabolic process. 
That being said, you can work your way up to cellular amino acid metabolism.

Antonialock:
ok cheers

ValWood:
I think this is a bit strange though.

If "cellular amino acid metabolism" is a parent, then it must be logically correct to have the individual amino acid parents?

Although I had been wondering  during "Matrix" work if it was strictly correct to describe "aminoacylation" and "amino acid metabolism"

hdrabkin:
There is a slippery slope about this: do we want to describe everything that an amino acid is involved in as an amino acid metabolism. In this case the amino acid is not really catabolized; (the breakdown of complex molecules in living organisms to form simpler ones, together with the release of energy; destructive metabolism)
GO metabolism "The chemical reactions and pathways, including anabolism and catabolism, by which living organisms transform chemical substances." 
Anabolism: the synthesis of complex molecules in living organisms from simpler ones together with the storage of energy; constructive metabolism. Protein synthesis kind of fits here;  I'm not sure if people generally think of the role of storage of energy looking at a protein, protein is generally not used as an energy source unless the body has no other option.
Not sure what you mean by 
if it was strictly correct to describe "aminoacylation" and "amino acid metabolism"
DId you mean as' amino acid metabolism'??

deustp01:
Jumping in here in favor of narrowly drawn process boundaries, with Harold: translation is not a kind of amino acid metabolism, and if tRNA aminoacylation belongs anywhere, I'd think it belongs there. That's also an argument, responding to Val's comment from two days ago, that tRNA aminoacylation should not be a child of cellular amino acid metabolism.

Here's where terms like "causally upstream of" stop being euphemisms for "we don't exactly know what's going on" and have real logical power. Absent amino acid synthesis or uptake mediated by process one, aminoacylation cannot happen. The two distinct, non-overlapping processes are related in a precise way.

[...]

PR #32551, opened 2026-08-28
Obsolete 43 amino-acid-specific tRNA aminoacylation terms, #15375

raymond91125:
[Written by Claude]

Step 2 of the recommendation on #15375 ([comment](https://github.com/geneontology/go-ontology/issues/15375#issuecomment-5431742836)). Obsoletes 43 amino-acid-specific tRNA aminoacylation terms.

**Please read the "Approval status" section before merging.** This has not been signed off by @pgaudet or @ValWood, and one term on the list is contested.

## What changed

43 terms obsoleted. Each restates an existing molecular function with an exact 1:1 counterpart already in GO, so the BP term adds nothing the MF does not already state. Every obsolete stanza names its MF counterpart in the `comment:`.

`replaced_by` is compartment-aware rather than collapsing everything onto `GO:0043039`:

| | count | `replaced_by` |
|---|---|---|
| general (non-mitochondrial) | 22 | `GO:0006418` tRNA aminoacylation for protein translation |
| contested | 1 | `GO:0070681` — `consider` only, no `replaced_by` (see review fixes) |
| mitochondrial | 20 | `GO:0070127` tRNA aminoacylation for mitochondrial protein translation |

The tracked plan said `replaced_by: GO:0043039` for all of them. Splitting by compartment costs nothing and preserves the **mitochondrial** distinction that a single rollup would have destroyed. Note `GO:0006418` is the general parent, not a cytosolic term, so no cytosolic distinction is preserved by the other row — neither option retains one, since GO has no cytosolic tRNA charging term.

Surviving branch:

```
GO:0043039 tRNA charging
├─ GO:0006418 tRNA aminoacylation for protein translation
│   └─ GO:0070127 ...for mitochondrial protein translation
└─ GO:0043040 ...for nonribosomal peptide biosynthesis   [→ #30872]
```

## Axioms removed

Per `/term-obsoletion`, obsolete stanzas retain no logical axioms:

- **42 `is_a`**, including **20 `is_a: onlyin:2759`** taxon constraints on the mitochondrial terms
- **42 `intersection_of`** lines — 21 logical definitions, all of the form *&lt;cytosolic term&gt; + `occurs_in` &lt;compartment&gt;* (the 20 mitochondrial terms plus `GO:0061475`)
- **`xref: MetaCyc:PWY-5921`** on `GO:0070681`. The mapping retires with the term; that identifier is retained as a definition dbxref, which is provenance for the definition text rather than a live mapping. It appears nowhere else in the ontology.

None of the 43 had synonyms, so nothing was lost there.

## Two axioms deliberately not migrated

**`GO:0006434 has_part GO:0001717`** (conversion of seryl-tRNAsec to selenocys-tRNAsec) is **dropped, not moved** onto `GO:0006418`. Moving it would assert that *every* tRNA aminoacylation for protein translation has the selenocysteine conversion as a part, which is false — the same universal-quantification error that review caught on `part_of GO:0160307` in #32541. `GO:0001717` keeps its own `is_a GO:0019988 charged-tRNA amino acid modification` and is not orphaned. Flagging it because it is the one genuinely lost axiom in this PR.

**No taxon constraints transferred.** `GO:0070127` is already correctly constrained through its own hierarchy, so re-asserting `onlyin:2759` would be redundant.

## Files deliberately left alone

- `src/ontology/ld.txt` contains 20 stale `intersection_of` references to these terms. It is a 2022 one-off (commit `d4ebb79d1`, message "Test"), is not referenced anywhere in the Makefile, and is not consumed by any build. Cleaning it up is out of scope here, but it is probably dead weight worth deleting separately.
- `src/ontology/go-idranges.txt` lists 41 of the IDs. It is the historical ID-minting ledger recording who minted each ID; obsoletion does not change that history.

## Annotation and mapping impact

Tracked on geneontology/go-annotation#6525. Of **251** non-IEA annotations reviewed, **204 already carry the paired MF from the same reference** and need no migration. Per-annotation recommendations with quotes from each paper are in column A of the review sheet.

Six annotations should **not** move to a ligase MF at all and are flagged there for their groups: `gtpbp3` (ZFIN, 6 rows — a tRNA-modifying enzyme, not a synthetase), `Tb11.01.7110` (GeneDB — a formyltransferase), `DARS2` (UniProt — wrong substrate), and the Arc1p/MES1/GUS1 subunit propagation (ComplexPortal).

**170 mappings** point at these terms — InterPro2GO 63, UniRule 62, HAMAP 44, MetaCyc 1 — and will need retargeting to the MF terms. Mapping contacts are assigned on #6525.

## Approval status

Two things a reviewer should weigh rather than assume:

1. **Not signed off.** @pgaudet stated the intent in [Feb 2025](https://github.com/geneontology/go-ontology/issues/15375#issuecomment-2669074122) ("I think we need to obsolete the entire branches, as they are MFs") and @ValWood restated it in Sept 2025 and Jan 2026, so the direction is well established. But neither has approved this specific 43-term list or the compartment-aware `replaced_by` split.

2. **`GO:0070681` is contested.** @ValWood [flagged it as valid](https://github.com/geneontology/go-ontology/issues/15375#issuecomment-2669422582). It is included here only on the precedent of #31932, which obsoleted its exact structural twin `GO:0070680` as "a specific pathway variant, which is out of scope for GO". Keeping one and not the other is hard to defend, but if @ValWood prefers to keep it, drop that one stanza and #31932 should be revisited too. It carries 770 annotations, 14 experimental — the largest single block in this PR.

## Checklist

- [x] **PLAN** — implements step 2 of the recommendation on #15375
- [x] **PRE-VALIDATION** — branched from `origin/master` at d641597cc; `travis_test` clean before and after
- [x] **RESEARCH** — N/A; no references introduced. The 12 papers behind the contested annotations were read and are quoted on #6525
- [x] **TERM-SEARCH** — all 43 BP→MF pairs derived from `go-edit.obo` programmatically, not hand-typed; 0 unmapped
- [x] **DESIGN-PATTERNS** — N/A; no new terms, no logical definitions added
- [x] **EDITS** — `obo-checkout.pl` / `obo-checkin.pl` via `terms/`; edit file never touched directly
- [x] **RELATIONSHIPS** — all logical axioms removed from obsolete stanzas; the one non-migratable axiom is documented above
- [x] **SPECIALIZED-EDITS** — `/term-obsoletion` followed: `obsolete` name prefix, `OBSOLETE.` def prefix, reason in `comment:`, `term_tracker_item`, `replaced_by`, no residual axioms; annotation impact assessed
- [x] **METADATA** — no `created_by`/`creation_date` added; the three pre-existing ones retained
- [x] **AUTOMATED-VALIDATION** — `make travis_test` exit 0; 16/16 SPARQL rules including `obsolete-reference-violation` (no live term references any of the 43) and `missing_superclass-violation`; `change-report.txt` empty
- [x] **REFERENCE-VALIDATION** — no PMIDs, ORCIDs or GOC IDs added
- [x] **CHANGES-COMMITTED** — only `src/ontology/go-edit.obo`


raymond91125:
[Written by Claude]

Both IMPORTANT findings accepted and fixed in d5e49f3b0. Point by point.

## IMPORTANT 1 — GO:0070148 cites the wrong MF — **accepted, fixed**

Correct, and the sting is that I had the evidence in hand. PMID:19805282 was one of the twelve papers I read for the annotation review on geneontology/go-annotation#6525, and it says it outright:

> "no mitochondrial homolog of glutaminyl-tRNA synthetase (GlnRS) has been identified in mammalian genomes, implying that Gln-tRNA(Gln) is synthesized via an indirect pathway in the mammalian mitochondria"

I generated all 43 comments from a programmatic name match (`glutaminyl-` → `glutamine-tRNA ligase activity`) and never re-checked the biology on the one term where the name match is the wrong answer. The PR body's claim that programmatic derivation avoided transposition errors was true but beside the point here.

The comment now names GO:0050567 as the counterpart in most eukaryotic mitochondria, with GO:0004819 applying only where a mitochondrial GlnRS is present. `replaced_by: GO:0070127` is unchanged and remains correct.

## IMPORTANT 2 — GO:0070681 `replaced_by` points at a sibling — **accepted, fixed**

Confirmed against `origin/master`: both `GO:0070681` and `GO:0006418` were `is_a: GO:0043039`. Siblings, so `replaced_by` was not a generalisation and would have made 770 annotations more specific than their evidence supports on automatic remap.

Took your preferred option rather than the safe fallback:

```
consider: GO:0043039
consider: GO:0050567
```

with no `replaced_by`. Your third sub-point is what decided it — GatCAB subunits act on an already-charged tRNA, so a chunk of those annotations plausibly belong under `GO:0019988 charged-tRNA amino acid modification` rather than under tRNA charging at all. No single target is safe, so blocking auto-migration is right.

I also moved the reasoning into the stanza's `comment:`, since your observation that `GO:0050567` "is nowhere machine-readable" applied equally to the rationale sitting only in a commit message.

## SUGGESTION 3 — split GO:0070681 out — **not applied, and it is not my call**

I agree with the reasoning: it is the only contested stanza, the only one with a different obsoletion reason, and the only one that carried a finding. Holding 42 clean stanzas behind one judgment call is a poor trade.

But the 43-term scope was set by the requester, and narrowing what was asked for is their decision rather than mine. @raymond91125 — say the word and I will split `GO:0070681` into its own PR, leaving 42 here. With the `consider`-only fix above it is no longer *harmful* to ship it in this batch; it is now purely a question of whether you want the 42 to move independently of @ValWood's objection.

## On your framing correction

Fair, and I have fixed it in the PR body. `GO:0006418` is the general parent, not a cytosolic term, so calling that row "cytosolic / general" overstated what the split preserves. It preserves the **mitochondrial** distinction, which is the one GO actually has a term for. The alternative — a flat rollup to `GO:0043039` — would have discarded that too, which is the comparison that matters.

## Validation

`make travis_test` in the pinned ODK image: exit 0, 16/16 SPARQL rules pass, `change-report.txt` empty.

Final tally across the 43: 22 × `replaced_by: GO:0006418`, 20 × `replaced_by: GO:0070127`, 1 × `consider` only.

Issue #20574, opened 2020-12-15
Evaluate logical definitions using 'process_has_causal_agent ' 


pgaudet:
Discussion of #17613 on the Ontology editors call highlighted that the relation 'process_has_causal_agent ' is probably incorrect. This relation was also questioned here: #20258.  

 - [ ]	intersection_of: process_has_causal_agent CHEBI:16480 ! nitric oxide
 - [ ]	intersection_of: process_has_causal_agent CHEBI:16670 ! peptide
 - [ ]	intersection_of: process_has_causal_agent CHEBI:16838 ! polyphosphate
 - [ ]	intersection_of: process_has_causal_agent CHEBI:17245 ! carbon monoxide
 - [ ]	intersection_of: process_has_causal_agent CHEBI:18059 ! lipid
 - [ ]	intersection_of: process_has_causal_agent CHEBI:33697 ! ribonucleic acid
 - [ ]	intersection_of: process_has_causal_agent CHEBI:48433 ! angiotensin
 - [ ]	intersection_of: process_has_causal_agent CHEBI:67194 ! cannabinoid
 - [ ]	intersection_of: process_has_causal_agent CL:0000084 ! T cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000097 ! mast cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000129 ! microglial cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000233 ! platelet
 - [ ]	intersection_of: process_has_causal_agent CL:0000236 ! B cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000451 ! dendritic cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000542 ! lymphocyte
 - [ ]	intersection_of: process_has_causal_agent CL:0000545 ! T-helper 1 cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000546 ! T-helper 2 cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000623 ! natural killer cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000624 ! CD4-positive, alpha-beta T cell
 - [ ]	intersection_of: process_has_causal_agent CL:0000738 ! leukocyte
 - [ ]	intersection_of: process_has_causal_agent CL:0000766 ! myeloid leukocyte
 - [ ]	intersection_of: process_has_causal_agent CL:0000767 ! basophil
[...]

Issue #22569, opened 2021-12-13
Fix LD of some transport terms


pgaudet:
Discussion with @balhoff and @cmungall about logical definitions: 

added logical definition: 
**exocytosis**
+intersection_of: GO:0006810 ! transport
+intersection_of: process_has_causal_agent GO:0070382 ! exocytic vesicle

**exosomal secretion**
+intersection_of: GO:0006887 ! exocytosis
+intersection_of: transports_or_maintains_localization_of GO:0070062 ! extracellular exosome

Thanks, Pascale

PR #32559, opened 2026-09-01
Drop the redundant process_has_causal_agent relationship on GO:0098876

raymond91125:
**[Written by Claude]**

Spotted while surveying `process_has_causal_agent` usage for #20574. One line.

`GO:0098876 vesicle-mediated transport to the plasma membrane` carried the same differentia twice:

```
intersection_of: GO:0006810 ! transport
intersection_of: has_target_end_location GO:0005886 ! plasma membrane
intersection_of: process_has_causal_agent GO:0031982 ! vesicle
relationship: ends_with GO:0006887 ! exocytosis
relationship: process_has_causal_agent GO:0031982 ! vesicle   <- removed
```

The equivalence axiom entails the subclass assertion, so the `relationship:` line was already implied. The stanza was also inconsistent with itself — `has_target_end_location` sits in the logical definition with no duplicate relationship.

`relationship: ends_with GO:0006887` is untouched: it is a genuine additional assertion, not part of the logical definition.

## Why this is a defect and not the house style

Duplicating a differentia as a relationship *is* a real GO pattern - 109 stanzas do it - so this needed checking before deleting. The breakdown by relation:

| relation | duplicated lines |
|---|---|
| `part_of` | 66 |
| `regulates` | 26 |
| `positively_regulates` | 5 |
| `negatively_regulates` | 4 |
| `has_primary_input` | 4 |
| `capable_of` | 2 |
| `has_part`, `has_target_end_location`, `has_target_start_location`, `results_in_development_of` | 1 each |
| **`process_has_causal_agent`** | **1** |

**Correction to an earlier version of this description.** It claimed the 101 `part_of`/`regulates` duplicates are load-bearing because `go-simple-filtered.ofn` (`Makefile:418`) drops `EquivalentClasses`, so the duplicate is what preserves the link in `go-basic` - and that this instance is safe because `process_has_causal_agent` is not in `GO_BASIC_RELATIONS`. Both halves were wrong on mechanism, as the review points out. Verified against the Makefile:

1. **`process_has_causal_agent` is stripped earlier than `go-basic`.** `$(ONT)-pre.ofn` (`Makefile:397`) runs `remove $(addprefix --term ,$(GO_MAIN_RELATIONS)) --select complement --select object-properties --signature true --trim true`, keeping only `GO_MAIN_RELATIONS` (`Makefile:64`). The relation is in neither that set nor `GO_BASIC_RELATIONS`, so it never reaches `go.owl`, let alone `go-basic`.
2. **The duplicates do not preserve the `part_of`/`regulates` links either.** `reasoned.ofn` (`Makefile:331`) runs `relax` before any of this, rewriting `EquivalentClasses(A, B and R some C)` into the corresponding `SubClassOf` axioms. By the time `Makefile:418` removes `EquivalentClasses`, the relaxed axioms already carry those links, identical to what a `relationship:` line produces.

So all 109 duplicates are inert in the release products, not just this one. The conclusion is unchanged and if anything stronger - the deleted line bought nothing anywhere - but the test I originally gave was wrong, and anyone reusing it on the other 108 would get the wrong answer. Corrected here rather than left to mislead.

## Verification

No logical change by construction: `EquivalentClasses(X, ...  ⊓ ∃process_has_causal_agent.vesicle)` entails `SubClassOf(X, ∃process_has_causal_agent.vesicle)`.

I did not run the before/after ELK subclass-pair comparison used on the other #20574 PRs, because it would not be informative here — that query filters to named-class subsumptions (`isIRI(?p)`), and this axiom's target is an anonymous restriction, so these triples never appear in it. A 0/0 result would have proved nothing. `make travis_test` does run the full reasoner and passes: 16/16 SPARQL rules, `change-report.txt` empty.

## Checklist

- [x] PLAN - single redundant axiom, scope confirmed against the 109-stanza pattern before deleting
- [x] PRE-VALIDATION - master validates; branch cut from current master
- [ ] RESEARCH - N/A, no biological claim
- [x] TERM-SEARCH - full scan for identical `intersection_of` / `relationship` pairs
- [x] DESIGN-PATTERNS - checked against `GO_BASIC_RELATIONS` and the `go-basic.ofn` recipe
- [x] EDITS - `obo-checkout.pl` / `obo-checkin.pl`
- [x] RELATIONSHIPS - logical definition unchanged; `ends_with` retained
- [ ] SPECIALIZED-EDITS - N/A
- [x] METADATA - nothing added or altered
- [x] AUTOMATED-VALIDATION - `make travis_test`, exit 0
- [ ] REFERENCE-VALIDATION - N/A, no references introduced
- [x] CHANGES-COMMITTED - `src/ontology/go-edit.obo` only

## Related

Not part of the group A work in #32553 / #32554, which are merged. The remaining `process_has_causal_agent` cleanup on #20574 is the 68 non-CL logical definitions and the 8 other `relationship:` assertions — one of which, `GO:0007610 behavior -> UBERON:0001016 nervous system`, looks questionable on biology since it excludes plant and unicellular behaviour. Not addressed here.


raymond91125:
**[Written by Claude]**

Both corrections accepted, and I verified them against the Makefile before editing rather than taking them on trust:

1. **`Makefile:397`** - `$(ONT)-pre.ofn` runs `remove $(addprefix --term ,$(GO_MAIN_RELATIONS)) --select complement --select object-properties --signature true --trim true`, so every object property outside `GO_MAIN_RELATIONS` (`Makefile:64`) is gone before `go.owl`. `process_has_causal_agent` is in neither that set nor `GO_BASIC_RELATIONS`. You are right that I cited a gate one step too late.

2. **`Makefile:331`** - `reasoned.ofn` runs `relax` ahead of everything downstream, so the `part_of`/`regulates` links are already carried by relaxed `SubClassOf` axioms by the time `Makefile:418` drops `EquivalentClasses`. The 101 duplicates I called load-bearing are inert too.

The PR description is corrected, with the original claim quoted and marked as wrong rather than silently replaced - your point that the test would misfire on the other 108 is exactly why it needed fixing in place rather than just dropping.

The conclusion is unchanged and slightly stronger: the deleted line bought nothing in any release product, by an earlier gate than I described.

On your closing note - that retiring `intersection_of: process_has_causal_agent GO:0031982` from this stanza would now take the causal-agent link with it, rather than leaving a `relationship:` behind: noted, and worth stating where #20574 has actually landed on that. Group B was measured and the relation is being **kept** there (demoting it produced 52 false subsumptions, including `trans-synaptic signaling by BDNF` under `by soluble gas`). Group C, which `GO:0098876` belongs to, was triaged as no-action, and the exocytosis axiom in this family was added deliberately in #22569. So the likely outcome for this stanza is that the logical definition stays as it is, and the question does not arise - but you are right that it should be a decision rather than a side effect, and it is recorded on the issue.

Issue #31923, opened 2026-04-20
Textual definition update: GO:0045022  early endosome to late endosome transport (minor)


ValWood:
GO:0045022  early endosome to late endosome transport
Definition (GO:0045022 GONUTS page)
The directed movement of substances, in membrane-bounded vesicles, from the early sorting endosomes to the late sorting endosomes; transport occurs along microtubules and can be experimentally blocked with microtubule-depolymerizing drugs. PMID:29980602

Not always microtubule dependent (in fission yeast it's actin dependent)
 remove gloss:
transport occurs along microtubules and can be experimentally blocked with microtubule-depolymerizing drugs

ValWood:
@dragon-ai-agent  please update definition

dragon-ai-agent:
🤖 Working on it...

Follow along: [View workflow run](https://github.com/geneontology/go-ontology/actions/runs/24725790991)

*— @dragon-ai-agent*

dragon-ai-agent:
Definition updated on GO:0045022 to drop the microtubule-dependence clause. New text:

> The directed movement of substances, in membrane-bounded vesicles, from the early sorting endosomes to the late sorting endosomes.

Rationale: as you noted, the mechanism isn't universal — fission yeast uses actin rather than microtubules for this transport, so the gloss was over-specific. Logical axioms, relationships, and synonyms left untouched.

PR: #31938

---
🤖 **Generated by @dragon-ai-agent**
- Model: `claude-opus-4-7`
- Agent harness: claude-code
- Triggered by: @ValWood
- Run: [View workflow run](https://github.com/geneontology/go-ontology/actions/runs/24725790991)

ValWood:
@dragon-ai-agent please 
always add the term tracker url

dragon-ai-agent:
🤖 Working on it...

Follow along: [View workflow run](https://github.com/geneontology/go-ontology/actions/runs/24730330006)

*— @dragon-ai-agent*

dragon-ai-agent:
Added `term_tracker_item` pointing to this issue on GO:0045022 (alongside the existing link to the original creation issue #26386). Pushed to the same branch, so PR #31938 now contains both the definition change and the new tracker.

I've also saved this as a durable preference so future edits append a `term_tracker_item` for the driving issue by default.

---
🤖 **Generated by @dragon-ai-agent**
- Model: `claude-opus-4-7`
- Agent harness: claude-code
- Triggered by: @ValWood
[...]

Issue #21716, opened 2021-06-28
Term merge: [MVB terms, difference?]


ValWood:
**Term to be merged: Provide GO ID and term label**

GO:0032511 late endosome to vacuole transport via multivesicular body sorting pathway
defined
The directed movement of substances from endosomes to vacuoles by a pathway in which molecules are sorted into multivesicular bodies, which then fuse with the vacuole.

and 
GO:0043328
protein transport to vacuole involved in ubiquitin-dependent protein catabolic process via the multivesicular body sorting pathway 

**Reason for merge**

I don't see any difference between these 2 terms?

mah11:
The difference is clear: GO:0043328 is specific for proteins, whereas GO:0032511 is not. Whether that distinction *needs* to be made is a whole other question (I have no preference whether the terms are merged or not).

PR #32630, opened 2026-09-17
Connect ILV formation to MVB assembly and endosome maturation (#31770)

### LINKED ISSUE #31770: endosome transport via multivesicular body sorting pathway/GO:0045022early endosome to late endosome transport
CC 
@jobberkhout

Image

vesicle-mediated transport needs and overhaul.
I'm still working through documenting the required changes.
However, here is a potential quick fix improvement which will help with modelling.

endosome transport via multivesicular body sorting pathway
The directed movement of substances from endosomes to lysosomes or vacuoles by a pathway in which molecules are sorted into multivesicular bodies, which then fuse with the target compartment.

should be a descendant of 
GO:0045022 early endosome to late endosome transport
The directed movement of substances, in membrane-bounded vesicles, from the early sorting endosomes to the late sorting endosomes; ~transport occurs along microtubules and can be experimentally blocked with microtubule-depolymerizing drugs.~ delete this part of the definition, it isn't true for yeast (at least not all)

Justification
As an early endosome matures into a late endosome, it accumulates ILVs in its lumen — this is precisely what makes it a multivesicular body. So the MVB is the late endosome (or an intermediate stage en route to it), and the process of forming it is early-to-late endosome maturation.

I'm not even sure that 2 separate terms are justified, but we can look into that later. For now, I want to reconcile annotations in 2 different branches.

Note also there are quite a few unecessary grouping terms here "endosomal transport" for example

Also note that the organization branch is redundant with the transport branch and appears to be unnecessary, since the "organization" is a direct consequence of the transport process (largely used by synGO)

raymond91125:
Closes #31770 in part. Addresses #21716 only by analysis, not by edit -- see below.

> **Revised after review.** The first commit also asserted `GO:0036258 part_of GO:0045022`. That was over-asserted and has been reverted; see [Why the maturation link was dropped](#why-the-maturation-link-was-dropped). Two assertions remain.

## What was requested

#31770 asks for two things:

1. Remove the microtubule clause from the GO:0045022 definition. **Already done** under #31923; no action needed.
2. Make GO:0032509 *endosome transport via multivesicular body sorting pathway* a descendant of GO:0045022 *early endosome to late endosome transport*.

## Why the requested is_a was declined

GO:0045022 is a defined class with explicit endpoints:

```
intersection_of: GO:0016192 ! vesicle-mediated transport
intersection_of: has_target_start_location GO:0005769 ! early endosome
intersection_of: has_target_end_location  GO:0005770 ! late endosome
```

GO:0032509 ends at the lysosome/vacuole, not the late endosome. Its definition explicitly includes the terminal fusion, and GO:0061763 *multivesicular body-lysosome fusion* is asserted `part_of` its child GO:0032510. The assertion is therefore false at the top, and worse on inheritance: GO:0032511 and GO:0061764 inherit `has_target_start_location GO:0005770` from GO:0045324 and GO:1902774 respectively, so under the reasoner they start at the late endosome and would acquire a second, contradictory start location.

ELK would not catch this -- GO asserts no disjointness over these locations -- so nothing flags it but a reader. That is precisely why it should not go in.

## What was done

The requester's justification (ILV accumulation is what turns an early endosome into a late endosome/MVB) is sound, but applies one level down. GO:0070676 *intralumenal vesicle formation* was almost unconnected -- no link to MVB assembly, none to the MVB sorting pathway -- which is why annotations split across the two branches.

| term | added | rationale |
|---|---|---|
| GO:0070676 | `relationship: part_of GO:0036258` | ILV formation is the mechanism of MVB biogenesis |
| GO:0071985 | `relationship: has_part GO:0070676` | the pathway by definition involves cargo entry into luminal vesicles |

Each term also received `property_value: term_tracker_item ".../issues/31770" xsd:anyURI`. No `created_by`/`creation_date` -- both are pre-existing terms.

The second is deliberately `has_part` rather than the `part_of` originally floated on the issue. ILV formation also feeds extracellular exosome biogenesis (GO:0097734, which carries "ILV assembly" and "intraluminal vesicle assembly" as RELATED synonyms), where the MVB fuses with the plasma membrane and cargo is secreted rather than degraded. `GO:0070676 part_of GO:0071985` would place those events inside a pathway whose definition ends in lysosomal/vacuolar degradation. `has_part` states only that every MVB sorting pathway has an ILV formation part, which is true and is the same pattern GO:0043162 already uses.

## Why the maturation link was dropped

The first commit also asserted `GO:0036258 part_of GO:0045022`, to give the MVB branch a path to endosome maturation. Reverted in the second commit, on review. `part_of` is declared `is_transitive: true` in this ontology, so combined with `GO:0070676 part_of GO:0036258` it entailed `GO:0070676 part_of GO:0045022` -- putting exosome-destined ILV formation inside an early-endosome-to-late-endosome transport. That is the same over-assertion the `has_part` inversion above avoids, re-entering one level down by transitivity. GO:0036258 also covers MVB biogenesis generally rather than only the first invagination, so continued ILV budding into an already-late endosome is in scope, and for those the EE-to-LE transition has already completed.

Underneath it is an unresolved question: GO:0045022's asserted genus is `GO:0016192 vesicle-mediated transport`, but it carries "endosome maturation" as a RELATED synonym. Maturation and vesicular shuttling are competing accounts of the same transition, and no sound axiom connects MVB assembly to GO:0045022 in either direction until that is settled. That belongs in the vesicle-transport overhaul #31770 is part of, not here.

**Consequence for the issue:** the two remaining assertions connect ILV formation to MVB assembly and to the MVB sorting pathway, but do **not** reach GO:0045022. The annotation reconciliation #31770 actually asks for is therefore still open, blocked on the GO:0045022 question.

## Reasoner effect

ELK before/after over the whole edit file -- the two assertions, plus taxon-constraint propagation:

- `onlyin:2759` (Eukaryota) newly inferred on GO:0070676, GO:0071985, GO:0043162, GO:1905365 -- propagated from GO:0036258, which already carried it, along the `part_of`/`has_part`/`regulates` chain.
- GO:0061763 tightened from `onlyin:131567` (cellular organisms) to `onlyin:2759`.

All five are MVB/ILV processes; eukaryote-only is correct for each. No unsatisfiable classes, no other entailment changes. The propagation is unchanged by the revert, since its source is GO:0036258's pre-existing constraint reached via the surviving `part_of`.

Checked against the annotation corpus, since a narrowed constraint can invalidate previously legal annotations: GOLR reports 58 distinct annotations under these terms outside Eukaryota -- 30 *Trichomonas vaginalis* (NCBITaxon:412133) and 28 *Klebsormidium nitens* (NCBITaxon:105231). Both are eukaryotes, so this is a `taxon_subset_closure` indexing gap rather than a real conflict. Nothing is invalidated.

## Deferred: the redundancy in this branch

Not in this PR, flagged for a separate proposal. The branch repeatedly carries differentiae already entailed by the parent's definition. GO:0071985 reads *"a vesicle-mediated transport process in which transmembrane proteins are ubiquitylated to facilitate their entry into luminal vesicles of MVBs; upon subsequent fusion of MVBs with lysosomes or vacuoles, the cargo proteins are degraded"* -- so protein/ubiquitin scoping is inherited throughout. Direct annotation counts (GOLR, experimental in parentheses):

- **GO:0071985** 608 (30) vs **GO:0032509** 810 (28) -- same process, near-identical definitions. GO:0071985 has exactly one asserted child (GO:0032509) and one other use (`has_part` from GO:0043162). Merge candidate; suggest keeping GO:0071985.
- **GO:0043328** 1435 (33) vs **GO:0032511** 1118 (36) -- this is #21716, which was closed as completed on 2026-09-15 and folded into #31770, but no ontology change ever accompanied it: both terms are still present, unmerged, with no `alt_id`. Still live scope.
- The lysosome/vacuole split (GO:0032510 vs GO:0032511, GO:0061764) is the species-specific duplication. Real, but much larger.

## Checklist

- [x] **PLAN** -- issue and comments read, including the #21716 thread folded into it; intent clear
- [x] **PRE-VALIDATION** -- `go-edit.obo` clean at base before edits; perl OBO check exit 0
- [x] **RESEARCH** -- N/A beyond the ontology itself; no new literature references introduced
- [x] **TERM-SEARCH** -- `obo-grep.pl` over the MVB/endosome/ILV/exosome branches
- [x] **DESIGN-PATTERNS** -- no `intersection_of` added; only `relationship` tags, following prior art in this branch (GO:0043162 `has_part`, GO:0071971 `part_of`)
- [x] **EDITS** -- `obo-checkout.pl` / edit in `terms/` / `obo-checkin.pl`
- [x] **RELATIONSHIPS**
    - [x] no logical definitions added; existing ones untouched
    - [x] relationships conform to similar terms in the branch
    - [x] `is_a` not over-asserted -- no `is_a` added
    - [x] `part_of` vs `has_part` chosen per direction of entailment, and checked under `part_of` transitivity
- [x] **SPECIALIZED-EDITS** -- N/A (no obsoletion, chemistry, reaction, or hand-authored taxon constraint; TC changes here are all inferred)
- [x] **METADATA** -- `term_tracker_item` on both edited terms; no `created_by`/`creation_date` on pre-existing terms
- [ ] **AUTOMATED-VALIDATION** -- **partial locally; CI green on the previous head**
- [x] **REFERENCE-VALIDATION** -- N/A, no new PMIDs or xrefs
- [x] **CHANGES-COMMITTED** -- only `src/ontology/go-edit.obo`

### Validation actually run

The Docker socket is blocked in the editing sandbox for this session, so no ODK-image target could run locally. Run against host ROBOT 1.9.7 instead:

- `robot reason -r ELK` before and after, diffed -- results above; no unsatisfiable classes
- `robot verify` over all 12 `src/sparql/*-violation.sparql` QC queries -- **12/12 PASS**
- `perl ../util/check-obo-for-standard-release.pl --disable-isa-incomplete --disable-whitespace-check --xref-abbs GO.xrf_abbs go-edit.obo` -- clean, exit 0

**Not run locally**, deferred to CI: `datalog-check`, `qc-selftest`, `change-report.txt`, `check_taxon_constraint_idspaces`, `check_all_taxon_constraints_columns`, `chebi_pH_7_3_check`. All five checks passed on the previous head; worth confirming green again after the revert, particularly `change-report.txt` and the taxon-constraint column targets.


raymond91125:
[Written by Claude]

Thanks -- the IMPORTANT finding is correct and is fixed in 3c4470e2a. Responses to all three.

## IMPORTANT: `GO:0036258 part_of GO:0045022` -- accepted, reverted

Your point 2 is decisive and I verified it directly: `part_of` is declared `is_transitive: true` in `go-edit.obo`, so rows 1 and 2 together entailed `GO:0070676 part_of GO:0045022`. I inverted row 3 to `has_part` specifically to keep exosome-destined ILV formation out of a degradative pathway, then let it into a transport pathway through the back door one level down. That was inconsistent and you caught it.

Point 1 holds too: GO:0036258's definition scopes to MVB biogenesis generally, not the first invagination, so continued ILV budding into an already-late endosome is in scope and the EE-to-LE transition has completed for it.

Point 3 is the one I think matters most going forward. The genus/synonym split in GO:0045022 means no sound axiom connects MVB assembly to it in *either* direction -- `has_part` would fail under the vesicular-shuttling reading just as `part_of` fails under the maturation reading. I have said so in the PR body and flagged that the reconciliation #31770 actually asks for is now still open, blocked on that question rather than delivered by this PR. Better an honest gap than an axiom that is only true under one reading.

The GO:0036258 stanza is reverted in full, including its `term_tracker_item` -- no point stamping a term that no longer changes.

## SUGGESTION: the "start at the late endosome" clause -- disagreeing

This one I would rather keep. You read the text definitions, which are loose; the logical parents are not. GO:0032511 `is_a` GO:0045324, which carries `intersection_of: has_target_start_location GO:0005770`, and GO:0061764 `is_a` GO:1902774, same. So under the reasoner both do start at the late endosome, and the requested `is_a` would have given them a second, contradictory start location rather than merely an odd-looking one.

You are right that the def wording invites your reading -- GO:0032511's says "from endosomes" while its name and parent say late endosome. That is a real defect, but in the definition, not in the rebuttal. I have reworded the PR body to cite the inherited `has_target_start_location` explicitly so the next curator sees where the claim comes from.

## SUGGESTION: taxon narrowing vs the annotation corpus -- checked, clean

Good catch that CI would not cover this. GOLR reports 58 distinct annotations under the five terms outside Eukaryota:

| taxon | count |
|---|---|
| NCBITaxon:412133 *Trichomonas vaginalis* | 30 |
| NCBITaxon:105231 *Klebsormidium nitens* | 28 |

(Correcting figures I first posted as 71 total and 35/37: those summed per-term facet counts across overlapping `isa_partof_closure` sets and double-counted annotations subsumed under more than one of the five terms. The 58/30/28 above are from a single deduplicated query. The conclusion is unaffected. The commit message on 3c4470e2a still carries the 71; treat these as the correct figures.)

Both are eukaryotes, so these are a `taxon_subset_closure` indexing gap in GOLR rather than genuine violations. Nothing is invalidated by the narrowing. Recorded in the PR body.

Propagation is unchanged by the revert, since its source is GO:0036258's pre-existing `onlyin:2759` reached via the surviving `GO:0070676 part_of GO:0036258`.

Issue #32632, opened 2026-09-18
GO:0032456 endocytic recycling: label and definition assert endocytic origin, logical definition does not


raymond91125:
[Written by Claude]

Filed as follow-up to #31237 / #32631, at the AI reviewer's request, so this doesn't get buried in a merged PR body.

## Problem

GO:0032456 `endocytic recycling` asserts an endocytic origin in its **label** ("endocytic") and its **definition** ("returns internalized cargo"), but its logical definition requires no such thing:

```obo
intersection_of: GO:0016192 ! vesicle-mediated transport
intersection_of: has_target_start_location GO:0005768 ! endosome
intersection_of: has_target_end_location   GO:0005886 ! plasma membrane
```

Any vesicle-mediated transport from an endosome to the plasma membrane satisfies this, whether or not the cargo was ever at the cell surface. So **biosynthetic** delivery — newly synthesised material routed TGN → recycling endosome → plasma membrane — classifies as `endocytic recycling` by the reasoner, which is wrong by the term's own name.

GO already has the upstream leg of that route as a separate term: GO:0044795 `trans-Golgi network to recycling endosome transport`. Its downstream leg has nowhere to go but GO:0032456.

## Why it wasn't fixed in #32631

#31237 asked whether GO:0032456 and GO:0099638 `endosome to plasma membrane protein transport` should be merged. They should not; the apparent duplication came from a **cargo** restriction in GO:0032456's text definition that its axioms did not carry. #32631 fixed that half by loosening the text to match the axioms.

The origin half is the same defect class but needs the opposite remedy — tightening the axioms, not loosening the text — and the right axiomatisation isn't obvious. Mixing the two into one PR would have muddied both.

## Options

1. **Tighten the LD** so endocytic origin is required. Needs a relation that expresses "cargo was previously internalised"; there may not be a clean one in RO, and it would have to not break the ~136 EXP annotations currently on the term.
2. **Add a sibling** for biosynthetic endosome → plasma membrane delivery under GO:0016197 `endosomal transport`, leaving GO:0032456's LD as the union parent. Then GO:0032456's name/def would need to change, or a new cargo-agnostic parent inserted.
3. **Accept the looseness**, document it in a `comment:`, and treat the endocytic reading as a labelling convention rather than an assertion. Cheapest; leaves the reasoner able to misclassify.

[...]

Issue #31237, opened 2025-12-18
Term merge: GO:0032456  endocytic recycling and GO:0099638    endosome to plasma membrane protein transport


ValWood:
* **Term to be merged: Provide GO ID and term label**

GO:0032456  endocytic recycling and  GO:0099638    endosome to plasma membrane protein transport

The only difference between these terms is that GO:0099638 refers to proteins. However, the definition of

GO:0032456  endocytic recycling
also specifies proteins:
The directed movement of membrane-bounded vesicles from endosomes back to the plasma membrane, a trafficking pathway that promotes the recycling of internalized transmembrane proteins. 

* **Which term do you propose this term merge into? Provide GO ID and term label**

* **Reason for merge**

* **Reference(s)**
PMID:nnnnnn

* **Any other information**

pgaudet:
This paper https://pubmed.ncbi.nlm.nih.gov/23563491/
describes glucose and metal ion transport. So: either we need a sibling for non-protein substrates, if the endocytic pathway is different, or we need to merge (obsolete+ replace) and fix definitions to broaden the concept.

ValWood:
this is because of the endocytosis of transporters for glucose and copper. It's still protein  that is being endocytosed. Other branches of vesicle -mediated transport can apply to non-proteins, but  endocytic recycling only applies to proteins (and that is mirrored by the definition)

raymond91125:
<img width="1209" height="828" alt="Image" src="https://github.com/user-attachments/assets/144118cb-21e8-4400-be92-e55f7a1c6f8b" />

raymond91125:
[Written by Claude]

## Conclusion

**Do not merge.** GO:0099638 is already an inferred child of GO:0032456, not a duplicate. The cargo restriction is the child term's whole content, and it carries parentage GO:0032456 cannot: `protein localization to plasma membrane`, `establishment of protein localization to plasma membrane`, `intracellular protein transport`.

**Next step for this ticket:** tighten GO:0032456, whose text definition and logical definition disagree — the mismatch is what made the two terms look like duplicates.

- **Broaden the text definition** to match the cargo-agnostic LD and the sibling family: *"The directed movement of substances in membrane-bounded vesicles from endosomes back to the plasma membrane, a trafficking pathway that returns internalized cargo, including transmembrane proteins and membrane lipids, to the cell surface."* [PMID:15040445, PMID:16473635, PMID:23563491]
[...]

PR #32639, opened 2026-09-18
Obsolete GO:0099638 endosome to plasma membrane protein transport and its regulation children (#31237)

raymond91125:
Refs #31237 — deliberately **not** `Fixes`. The requester closes that ticket once they are satisfied.

## Summary

Obsoletes four terms, each with a direct `replaced_by`:

| obsoleted | replaced_by |
|---|---|
| GO:0099638 endosome to plasma membrane protein transport | GO:0032456 endocytic recycling |
| GO:1905749 regulation of ~ | GO:2001135 regulation of endocytic recycling |
| GO:1905750 negative regulation of ~ | GO:2001136 negative regulation of endocytic recycling |
| GO:1905751 positive regulation of ~ | GO:2001137 positive regulation of endocytic recycling |

**Why.** pgaudet on #31237: *"the single-child pattern is always suspicious. If we cannot justify creating a sibling for 'endosome to plasma membrane protein transport', then it should merge up."* GO:0099638 was the only cargo-restricted child of GO:0032456, no sibling for non-protein cargo could be justified, so the distinction was not earning a term. ELK already inferred `GO:0099638 is_a GO:0032456`.

An earlier revision of this PR argued instead from PMID:8509447 that protein and lipid recycle with indistinguishable kinetics, so protein cargo does not define a distinct process. That was withdrawn in `334472ef8`. The finding is real, but it is the wrong rationale here: it contradicts ValWood, who authorised the merge while holding endocytic recycling "should still be defined as protein only"; it cuts the wrong way, since if non-protein cargo also recycles then protein cargo is a well-formed *narrower* class, which is why ELK inferred `is_a` and not equivalence; and applied consistently it would obsolete GO:0043001 against GO:0006893, a pair this ontology deliberately keeps. The evidence is retained in `RESEARCH_31237_cargo.md` for the cargo-scope discussion, not used as the obsoletion reason.

Per current GO practice these are obsoletions with `replaced_by`, **not merges**: no `alt_id`, the stanzas remain, the IDs still resolve.

## The regulation triad is not a separate decision

Rewiring `regulates` onto GO:0032456 makes GO:1905749/50/51 inferred-equivalent to GO:2001135/6/7, which ROBOT rejects ("Only equivalent classes that have been asserted are allowed"). They obsolete with the process or not at all.

## Descendants reparent without manual intervention

Verified by reasoning the edited file with ELK, not assumed:

| term | lands at |
|---|---|
| GO:0099639 | GO:0032456, GO:0098877 — **retains** GO:0006886, GO:0061951, GO:0072659 |
| GO:0098887 | unchanged, under GO:0099639 |
| GO:0099152 | GO:2001135 — **retains** GO:0033157, GO:1903076, GO:1902473, GO:1902683 |

Reviewer note: **GO:0099152 is an inferred-only child of GO:1905749** with no asserted reference, so it does not appear in an `obo-grep.pl` of the edit file. It is the one thing easy to miss here. It comes through correctly.

All three survivors are in `goslim_synapse`; the slim is unaffected. The only edit to a surviving term is removing `is_a: GO:0099638` from GO:0099639, which retains `is_a: GO:0099637`.

## Known consequence

GO:0099638's 246 annotations leave `GO:0072659 protein localization to plasma membrane`, `GO:0061951 establishment of protein localization to plasma membrane` and `GO:0006886 intracellular protein transport`, since GO:0032456 carries no cargo restriction. The descendants above keep that parentage via their own protein-transport logical definitions, so the loss is confined to those 246. ValWood accepted this on the issue, noting the parentage could be re-added and that she expects the protein-localization terms to be obsoleted eventually.

## GO:0032456 is deliberately untouched

An earlier revision of this section claimed GO:0032456 was left internally inconsistent. That was a misreading and is withdrawn.

Its definition is: *"The directed movement of membrane-bounded vesicles from endosomes back to the plasma membrane, a trafficking pathway that promotes the recycling of internalized transmembrane proteins."* The genus clause — directed movement of membrane-bounded vesicles from endosomes to the plasma membrane — carries no cargo restriction. The trailing clause says what the pathway *promotes*; it does not state that protein is the only cargo. So the text definition and the cargo-agnostic logical definition agree, and there is no conflict for this PR to reconcile.

Separately, #32632 records a real and unrelated looseness: the label ("endocytic") and definition ("internalized") assert an origin the logical definition does not require, so biosynthetic TGN → recycling endosome → PM delivery classifies here. That is untouched by this PR.

## Annotation review

geneontology/go-annotation#6557 — 36 EXP annotations across 7 groups (UniProt 14, MGI 8, PomBase 7, SGD 3, FlyBase 2, RGD 1, TAIR 1), one assignee per group. No InterPro2GO/UniRule/UniProtKW mappings target any of the four terms.

## Checklist

- [x] **PLAN** — issue and all comments analysed; three curators converged on merge-up after my initial recommendation against it was rejected on its merits
- [x] **PRE-VALIDATION** — perl OBO check clean on unmodified `go-edit.obo`; ELK reasoning clean before edits
- [x] **RESEARCH** — `/research` used earlier in this issue; PMID:15040445, PMID:19696797, PMID:30194414 validated with `linkml-reference-validator` (3/3). PMID:9153255 checked against PubMed.
- [x] **TERM-SEARCH** — `obo-grep.pl` for all four terms, all referring terms, and the `has_target_start_location endosome` family
- [x] **DESIGN-PATTERNS** — sibling family and the GO:0006893/GO:0043001 precedent consulted directly in the edit file
- [x] **EDITS** — `obo-checkout.pl` / `obo-checkin.pl` via `terms/`; edit file not hand-edited
- [x] **RELATIONSHIPS** — all `is_a`/`intersection_of` stripped from the obsoleted stanzas; one `is_a` removed from GO:0099639
- [x] **SPECIALIZED-EDITS** — `/term-obsoletion` skill followed. No CHEBI, no RHEA/EC, no taxon constraints
- [x] **METADATA** — `name: obsolete <original>`, `def: "OBSOLETE. <original>"`, obsoletion reason in `comment:`, `term_tracker_item` for #31237, synonyms preserved (8), `namespace` retained, `created_by`/`creation_date` untouched
- [x] **AUTOMATED-VALIDATION** — **CI green**: `ontology_qc` (which runs `make travis_test`, including the Soufflé `datalog-check` obsoletion rules) passed, as did `diff_asserted_axioms` and `validate_yaml_and_metadata`. Not runnable locally — Docker's socket is blocked in the editing sandbox and `souffle` is not on the host. Locally verified anyway: `check-obo-for-standard-release.pl` (exit 0), 15 SPARQL QC rules at 0 violations each, ELK reason exit 0 with no unsatisfiable classes and 0 inferred parents on all four obsoleted terms.
- [x] **REFERENCE-VALIDATION** — no references are introduced into the ontology by this PR. PMID:8509447 was added in `f0726a5ed` and removed again in `334472ef8`; it was verified against PubMed while present.
- [x] **CHANGES-COMMITTED** — single commit, `src/ontology/go-edit.obo` only, 24 insertions / 19 deletions
- [x] **ISSUE-ALIGNMENT** — performs the merge-up requested by pgaudet and agreed by ValWood on #31237


raymond91125:
[Written by Claude]

Thorough review — the reparenting check worked by axiom subsumption rather than taking the PR's word for it, and catching the false equivalence claims was the right call. Response to each finding.

### IMPORTANT 2 — obsoletion comments assert things that are false — **accepted, fixed**

Correct on both counts, and this was my error. A term differing by a cargo restriction is not a duplicate, and regulating a strict subclass is not equivalent to regulating its superclass. Since these are permanent provenance, the wording mattered.

Fixed in `6a8879d9b`, then superseded by `f0726a5ed`, which replaces the procedural wording with the actual biological reason (see below). The three regulation comments now read "…was merged into GO:0032456 …, so this term was merged into GO:2001135/6/7", describing the decision rather than asserting a logical relation.

### IMPORTANT 1 — add the cargo restriction to GO:0032456 — **declined; the premise does not hold**

The finding rests on GO:0032456 having a text/logical-definition conflict to reconcile. Re-reading the definition, it does not:

> "The directed movement of membrane-bounded vesicles from endosomes back to the plasma membrane, a trafficking pathway that promotes the recycling of internalized transmembrane proteins."

The genus clause — directed movement of membrane-bounded vesicles from endosomes to the plasma membrane — carries no cargo restriction. The trailing clause says what the pathway *promotes*; it does not state that protein is the only cargo. Text and axioms agree, and the PR body section claiming otherwise was mine, was wrong, and has been withdrawn.

The stronger point is that the restriction would be biologically wrong. Mayor, Presley & Maxfield (PMID:8509447) tracked fluorescent lipid analogues against transferrin in CHO cells and found them to traverse the pathway "with indistinguishable kinetics", concluding that constitutive sorting of bulk membrane toward the recycling pathway "implies a geometric basis of sorting" — what recycles is whatever sits in the membrane, not what it is. C6-NBD-phosphatidylcholine and galactosylceramide follow the same route. Adding `has_primary_input PR:000000001` would assert that protein is always the primary input of endocytic recycling, which that experiment contradicts. It would also break the cargo-agnostic `has_target_start_location endosome` sibling family, and contradict GO:0055037 `recycling endosome`, whose own definition names lipids.

On the curatorial record: the reading of ValWood's "it could be added" as conditional acceptance is reasonable but not the only one, and deustp01's question about dragged-along ligands is still unanswered on #31237. The editor driving this PR has made the call to leave GO:0032456 alone. Cargo scope remains open and is the right subject for its own ticket, not a line added under time pressure here.

### Adopted from the review indirectly

`f0726a5ed` states the rationale the obsoletion actually rests on: protein and lipid travel the same pathway together, so protein cargo does not carve out a distinct process. Deliberately *not* claiming the mechanism can never distinguish cargo — a sequence-specific layer is superimposed on bulk flow and separable from it (PMID:8376460, PMID:9153255), reading a cytoplasmic motif only proteins have. All five supporting quotes validated against PubMed with `linkml-reference-validator`.

### SUGGESTION — labels as synonyms on the replacements — **open**

Not actioned; it is a curator's call and the editor has not ruled on it. Noting for the record that if taken, scope would be NARROW given GO:0032456 stays cargo-agnostic.

### Validation

`ontology_qc` has since gone green, so the Soufflé `datalog-check` obsoletion rules that could not run locally are covered. The checklist box is now ticked on that basis rather than on the local partial checks.

raymond91125:
[Written by Claude]

### IMPORTANT 1 — obsoletion rationale states the rejected argument — **accepted, applied in `334472ef8`**

Right on all three counts, and the second is the one that should have stopped me writing it in the first place.

The comment now reads:

> The reason for obsoletion is that this term unnecessarily specified protein cargo, and no biologically meaningful sibling for non-protein cargo could be justified, so it was merged up into GO:0032456 endocytic recycling.

On the three points:

- **It contradicted ValWood.** She authorised the merge while holding that endocytic recycling "should still be defined as protein only". Writing the opposite into permanent provenance was wrong.
- **It cut the wrong way.** If non-protein cargo also recycles, protein cargo is a well-formed *narrower* class — which is exactly why ELK inferred `is_a` and not equivalence. The premise argues the subclass is legitimate, not redundant. Self-undermining, and I had no business not noticing: I cited GO:0006893/GO:0043001 earlier in this same issue as precedent for *keeping* such a pair.
- **It was inconsistent with GO:0043001.** Applied evenly it obsoletes that term too. The ontology deliberately keeps the pair.

The bulk-flow evidence itself stands — PMID:8509447's kinetics are real — it is just not the reason this term went away. It is retained in `RESEARCH_31237_cargo.md` for the cargo-scope discussion rather than asserted as the obsoletion rationale. PMID:8509447 is consequently no longer in `go-edit.obo`, and this PR now introduces no references into the ontology at all.

One correction to the finding, offered only so the next reader is not misled: the idiom is spelled `unnecssarily` in the file (31 occurrences, missing an `e`). I initially concluded the idiom did not exist because I grepped the correct spelling — the finding quoted it accurately. The new comment uses the correct spelling rather than propagating the typo.

### IMPORTANT 2 — cargo restriction on GO:0032456 — **declined, editor's decision**

You are right that my previous rebuttal missed the point: I argued text/LD internal consistency, and the finding was about axiom transfer. Those are different claims and I conflated them.

On the finding as actually framed: the editor has decided not to specify protein-only for GO:0032456. That is a curatorial call about what the term means, not a mechanical consequence of the obsoletion, and it is theirs to make. The consequence — GO:0099638's 246 annotations leaving GO:0072659, GO:0061951 and GO:0006886 — is recorded in the PR body and in the obsoletion notice, so it is not shipping silently.

Worth noting the tension this creates with IMPORTANT 1, which both of us now agree on: the accepted rationale is a granularity judgement that makes no claim about cargo. Leaving GO:0032456 cargo-agnostic is consistent with that. Adding the restriction would re-introduce a cargo claim the curators have not settled — deustp01's question on #31237 is still unanswered.

### SUGGESTION — labels as synonyms on the replacements — **still the editor's call, not actioned**

Scope would be NARROW as shipped.

### Validation

All four checks were green on `f0726a5ed`, including `ontology_qc` and therefore the Soufflé `datalog-check` obsoletion rules. `334472ef8` changes one comment line, so CI is re-running against it.

